PR #16, opened 2019-02-08
Test DateTime and JSON fields

tomchristie:
Tests for non-primative field types.


tomchristie:
* `JSON` is currently an expected failure in both MySQL and PostgreSQL (for different reasons).
* `DateTime` is now working.

Issue #10, opened 2019-02-07
Iterator interface


tomchristie:
We should allow cursored results, using an explicit `iterate` interface:

```python
async for row in database.iterate(query):
    ...
```

tomchristie:
Closed via #21

PR #21, opened 2019-02-09
Iterator support (#10)

weargoggles:
Add `iterate()` to DatabaseSession and implementations, where the implementations return an async generator from the libraries' existing async-iterable cursors.


tomchristie:
Oh this is lovely. A little simpler than I'd expected too.
Excellent stuff! 🎉

PR #22, opened 2019-02-10
Add LICENSE.md

weargoggles:
Package isn't installable with Pipenv currently because this file is missing but mentioned in setup.py.

Copied from encode/starlette.


tomchristie:
👍 👍

PR #34, opened 2019-02-13
SQLite support

tomchristie:
* Connection pooling not yet implemented.
* Proper `executemany` support not yet implemented.
* JSON field not supported. Believe it should be in SQLAlchemy 1.3, but didn't work off-the-bat.
* One failing test case - not coercing datetime type


tomchristie:
See #36

Issue #48, opened 2019-02-25
fetch_one raises TypeError in SQLite backend


alviezhang:
Versions:

Python==3.7.2
databases==0.1.7
aiosqlite==0.9.0
sqlalchemy=1.2.18

I have a table `notes` in sqlite, I run code below:

```python
query = notes.select().where(notes.c.id == note_id)
the_note = await database.fetch_one(query)
```

When `notes` table doesn't have the row that id equals to note_id, it raises

```
...
File "/Users/xxx/.local/share/virtualenvs/rose-Xd4LdsnO/lib/python3.7/site-packages/databases/core.py", line 97, in fetch_one
  return await connection.fetch_one(query=query)
File "/Users/xxx/.local/share/virtualenvs/rose-Xd4LdsnO/lib/python3.7/site-packages/databases/core.py", line 163, in fetch_one
  return await self._connection.fetch_one(query=query)
File "/Users/xxx/.local/share/virtualenvs/rose-Xd4LdsnO/lib/python3.7/site-packages/databases/backends/sqlite.py", line 102, in fetch_one
  return RowProxy(metadata, row, metadata._processors, metadata._keymap)
```

It seems that `RowProxy` doesn't take an empty row as parameter. And we can fix it just add two line in `databases/backends/sqlite.py`

```python
        async with self._connection.execute(query, args) as cursor:
            row = await cursor.fetchone()
+            if row is None:
+               return
            metadata = ResultMetaData(context, cursor.description)
            return RowProxy(metadata, row, metadata._processors, metadata._keymap)
```

And I found that MySQL backend has similar code in `databases/backends/mysql.py`:

```python
        try:
            await cursor.execute(query, args)
            row = await cursor.fetchone()
            metadata = ResultMetaData(context, cursor.description)
            return RowProxy(metadata, row, metadata._processors, metadata._keymap)
        finally:
            await cursor.close()
```

So I guess MySQL properly has some issue but I have not tested it.

alviezhang:
Here is a simple PR to fix it. https://github.com/encode/databases/pull/49

tomchristie:
Resolved in 0.1.8

PR #49, opened 2019-02-25
Fix SQLite fetch_one raises TypeError when row is None

alviezhang:
Fix https://github.com/encode/databases/issues/48


tomchristie:
Thanks for moving this forward!
We'd need test coverage for this to pass - do you reckon you'd be up for having a go at adding a test case too?

tomchristie:
Okay, acutally I got this covered in #50

tomchristie:
Pushed as 0.1.8 - thanks for flagging this up! 😄

PR #51, opened 2019-02-26
Update README.md

### LINKED ISSUE #47: More query examples for SQLAlchemy n00bs
Hi all,

I've started messing around with Databases to see how it could work with [Bocadillo](https://github.com/bocadilloproject/bocadillo). It looks quite promising, but there are a few issues I encountered while getting started.

Being not very familiar with SQLAlchemy, I had quite a hard time finding resources to learn the syntax mentioned in the README (SQLAlchemy Core). I used SQLAlchemy with Flask in the past but I think it had a different syntax (e.g. `db.commit()`, etc) which doesn't seem to be the one used here.

I also found that the docs for [SQLAlchemy Core](https://docs.sqlalchemy.org/en/latest/core/) the README links to are quite cryptic for new starters.

Do you think there could be more code examples? Especially about the following:

- How to perform a filter using a `WHERE` clause: `notes.select().where(id=1)`?
- Error handling, e.g. what happens when trying to retrieve a row that doesn't exist
- Accessing the ID (and perhaps full data) of the last inserted row

For what it's worth, I'd be glad to help out with this — does anyone have good introductory resources to SQLAlchemy Core aside from the official docs?

florimondmanca:
Fixes #47 (as far as I'm concerned).

- Document return types of `.execute()`, `.execute_many()`, `.fetch_all()` and `.fetch_one()`.
- Add link to SQLAlchemy Core tutorial.
- Add comments with file names and import statement for `notes`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ Driver support is providing using one of [asyncpg][asyncpg], [aiomysql][aiomysql
 Declare your tables using SQLAlchemy:
 
 ```python
+# tables.py
 import sqlalchemy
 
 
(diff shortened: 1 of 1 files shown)


tomchristie:
Thanks! I'd strongly prefer not to use type annotations in the example, or to split it apart as `tables.py`, `main.py`. We could defo take the tutorial link change as it is. The return types probably want to be worked into the documentation instead.

florimondmanca:
Okay — so should I reopen a PR just with the updated link to the tutorial?

tomchristie:
Yeah that's probably a good plan. 👍

PR #54, opened 2019-02-26
Fixes #53 Reformat debug log for queries

### LINKED ISSUE #53: Debug log entries for queries are wrong formatted
There is a debug log entry for each query within all backends, recording the query itself and its arguments.

The code is similar for Postgres, SQLite and MySQL backends:
```python
logger.debug(query, args)
```

That fails because the logging is trying to parse `args` and replace into `query`.

PeRDy:
It modifies the format of debug log entries for queries to fix #53. We could change the string template if it isn't appropriate.

### DIFF
--- a/databases/backends/mysql.py
+++ b/databases/backends/mysql.py
@@ -169,7 +169,7 @@ class MySQLConnection(ConnectionBackend):
             compiled._textual_ordered_columns,
         )
 
-        logger.debug(compiled.string, args)
+        logger.debug("Query: %s\nArgs: %s", compiled.string, args)
         return compiled.string, args, CompilationContext(execution_context)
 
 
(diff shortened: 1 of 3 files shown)


tomchristie:
Thanks!
We might want to just go with something else, but let's get in the fix first!

tomchristie:
Pushed to PyPI as 0.1.9

tomchristie:
Perhaps something like:

```python
f"{compiled.string!r}" if not args else f"{compiled.string!r}, {args!r}"
```

?

tomchristie:
Probably clearer what I mean when written as this I guess?

```python
repr(compiled.string) if not args else repr(compiled.string) + ", " + repr(args)
```

Issue #57, opened 2019-02-28
Expand documentation


tomchristie:
In particular:

* More examples of queries.
* Examples of using JSON field.
* Examples with joining tables.

tomchristie:
See also #31

mivade:
> Getting PK of inserted rows.

Is this currently possible?

mivade:
Digging into the code, it looks like it is in the same way as with SQLAlchemy (i.e., it gets returned when calling `execute` on an insert statement). Unfortunately this doesn't always return the primary key though. For example, with SQLite, the last "rowid" gets returned which corresponds to an integer primary key (in cases where the primary key is an integer) or the "undeclared" `ROWID` column: https://docs.python.org/3/library/sqlite3.html#sqlite3.Cursor.lastrowid

brainless:
Hey everyone, I am using databases for two of my products and would like to work on the documentation issue. I see this is still open. Since this GitHub issue  is very old I would like to know if the issue points are still valid. I will go through the documentation again today just to double check. Any pointers would be helpful as I am a bit hesitant to contribute to such an important project. Thanks

hp0404:
@brainless would be great tbh, currently stuck with joining tables :)

Pk13055:
Is there any update on this? This is a pretty good async library, but lack of documentation (+ generic name) makes it difficult to find/google anything complex..

PR #58, opened 2019-03-03
Support for `connection.raw_connection`

gvbgduh:
Adds ability to access the underlying driver's connection interface.

### DIFF
--- a/databases/backends/postgres.py
+++ b/databases/backends/postgres.py
@@ -96,8 +96,6 @@ class Record(Mapping):
     def __len__(self) -> int:
         return len(self._column_map)
 
-    # TODO (?) Nice __repr__
-
 
 class PostgresConnection(ConnectionBackend):
     def __init__(self, database: PostgresBackend, dialect: Dialect):
--- a/databases/backends/postgres.py
+++ b/databases/backends/postgres.py
@@ -74,18 +74,21 @@ class Record(Mapping):
             for idx, (column_name, _, _, datatype) in enumerate(self._result_columns)
         }
 
-    def __getitem__(self, key: str) -> typing.Any:
-        idx, datatype = self._column_map[key]
-        raw = self._row[idx]
+    def __getitem__(self, key: typing.Any) -> typing.Any:
         try:
-            processor = _result_processors[datatype]
-        except KeyError:
-            processor = datatype.result_processor(self._dialect, None)
-            _result_processors[datatype] = processor
+            idx, datatype = self._column_map[key]
+            raw = self._row[idx]
+            try:
+                processor = _result_processors[datatype]
+            except KeyError:
+                processor = datatype.result_processor(self._dialect, None)
+                _result_processors[datatype] = processor
 
-        if processor is not None:
-            return processor(raw)
-        return raw
+            if processor is not None:
+                return processor(raw)
+            return raw
+        except KeyError:
+            return self._row[key]
 
     def __iter__(self) -> typing.Iterator:
         return iter(self._column_map)
--- a/databases/backends/postgres.py
+++ b/databases/backends/postgres.py
@@ -7,7 +7,7 @@ from sqlalchemy.dialects.postgresql import pypostgresql
 from sqlalchemy.engine.interfaces import Dialect
 from sqlalchemy.sql import ClauseElement
 
-from databases.core import DatabaseURL
+from databases.core import DatabaseURL, NoBackendMethod
 from databases.interfaces import ConnectionBackend, DatabaseBackend, TransactionBackend
 
 logger = logging.getLogger("databases")
@@ -93,6 +93,8 @@ class Record(Mapping):
     def __len__(self) -> int:
         return len(self._column_map)
 
+    # TODO (?) Nice __repr__
+
 
 class PostgresConnection(ConnectionBackend):
     def __init__(self, database: PostgresBackend, dialect: Dialect):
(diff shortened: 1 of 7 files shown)


gvbgduh:
I followed 2 different approaches:
1. Pass the target method as `str` to the generic method like `raw_api_call`, so the driver can be called as `getattr(obj, str)`. While it saves some lines for implementation it can introduced unacceptable complexity for generalised solution. The source of problem that `mysql` and `sqlite` (and asyncpg in case of explicit cursor) require two actions to preform the query. It might also lead to a dead-end.
2. Just exposing the raw backend connection. It gives full flexibility and control, but requires more verbose implementation, which is probably ok for such cases.

If anything of these looks good I can get rid of redundant bits in the code and finish the rest of unittests. I'm also quite bad at naming.
Any feedback would be much appreciated.

gvbgduh:
Yeah, p. 2 (`expose_backend_connection`) looks somehow neat without interfering with existing, but might require some attention for releasing connections back to pool.

databases/backends/postgres.py
```diff
@@ -74,18 +74,21 @@ def __init__(self, row: tuple, result_columns: tuple, dialect: Dialect) -> None:
             for idx, (column_name, _, _, datatype) in enumerate(self._result_columns)
         }
 
-    def __getitem__(self, key: str) -> typing.Any:
-        idx, datatype = self._column_map[key]
-        raw = self._row[idx]
+    def __getitem__(self, key: typing.Any) -> typing.Any:
```
gvbgduh: Not happy with this implementation in general and especially passing `typing.Any`. I think it might be limited to `str` and `int` as a raw record's fields can be accessed by an int index or by the str key.

databases/backends/postgres.py
```diff
@@ -150,6 +150,10 @@ def __init__(self, database: PostgresBackend, dialect: Dialect):
         async for row in self._connection.cursor(query, *args):
             yield Record(row, result_columns, self._dialect)
 
+    async def raw_connection(self) -> asyncpg.connection.Connection:
```
tomchristie: This doesn't need to (shouldn't) be an async method. (It's not performing any I/O itself, just returning the handle)

I guess we should probably use a property too, rather than a method.

```python
@property
def raw_connection(self):
    ...
```

tomchristie: It's possible that the implementations ought to use a `def get_raw_connection()` method, and the developer facing interface should be a `def raw_connection` property, that calls `get_raw_connection()`

databases/core.py
```diff
@@ -111,6 +111,10 @@ def __init__(
             async for record in connection.iterate(query):
                 yield record
 
+    async def raw_connection(self) -> typing.Any:
+        async with self.connection() as connection:
+            return await connection.raw_connection()
```
tomchristie: This shouldn't be exposed at the top-level. Ie. You *have* to acquire the connection first, in order to access `.raw_connection`

databases/core.py
```diff
@@ -168,6 +172,9 @@ def __init__(self, backend: DatabaseBackend) -> None:
     async def execute_many(self, query: ClauseElement, values: list) -> None:
         await self._connection.execute_many(query, values)
 
+    async def raw_connection(self) -> typing.Any:
+        return await self._connection.raw_connection()
```
tomchristie: Let's go with...

```python
@property
def raw_connection(self) -> typing.Any:
    return self._connection.get_raw_connection()
```

tomchristie: Also let's move the implementation to the very end of the class. That layout feels more natural to me.

tests/test_databases.py
```diff
@@ -514,3 +514,95 @@ def run_sync(*args, **kwargs):
...
+        async with database.connection() as connection_1:
+            async with database.connection() as connection_2:
+                assert connection_1 is connection_2
+
+                raw_connection_0 = await database.raw_connection()
+                raw_connection_1 = await connection_1.raw_connection()
+                raw_connection_2 = await connection_2.raw_connection()
+
+                assert raw_connection_0 is raw_connection_1 is raw_connection_2
+                assert raw_connection_0 is connection_1._connection._connection
```
tomchristie: Using properties I think we'd really want to change this to something like this...

```python
        async with database.connection() as connection_1:
            async with database.connection() as connection_2:
                assert connection_1 is connection_2
                assert connection_1.raw_connection is connection_2.raw_connection
```

tests/test_databases.py
```diff
@@ -514,3 +514,95 @@ def run_sync(*args, **kwargs):
...
+@async_adapter
+async def test_queries_with_expose_backend_connection(database_url):
+    """
+    Replication of `execute()`, `execute_many()`, `fetch_all()``, and
+    `fetch_one()` using the raw driver interface.
+    """
+    async with Database(database_url) as database:
+        async with database.transaction(force_rollback=True):
+            # Get the raw connection
+            con = await database.raw_connection()
```
tomchristie: Let's use something like `connection`, `raw_connection`, or `driver` rather than the shorthand variable name.

databases/backends/postgres.py
gvbgduh: yeah, was thinking about the property as well, and naming would be more natural as well.

databases/core.py
gvbgduh: That makes sense and sets explicit flow.

gvbgduh: That also gives clarity. As connection already acquired (and it seems reasonable to avoid spawning the raw_connection from the database level acquiring a new one) we can expose already allocated connection object, that makes it a perfect case for the property.

databases/core.py
gvbgduh: Happy with layout, but considering the given context I think I'm looking forward to:

```
@property
def raw_connection(self) -> typing.Any:
    return self._connection.raw_connection
```

if there're no objections

tests/test_databases.py
gvbgduh: yes, that would be neat

databases/core.py
tomchristie: That's fine with me, yup. 👍

tomchristie:
Looks great! I don't *think* there's anything else to do on this part right? You happy for it to go in now?

gvbgduh:
Yeah, I'm quite happy now, can't think of anything else as well. Thanks for the feedback!

gvbgduh:
I'll prepare the documentation examples tomorrow, but within the 
https://github.com/encode/databases/issues/57

tomchristie:
@gvbgduh Makes sense - thanks.
Released now, as 0.1.10.

PR #60, opened 2019-03-06
Examples using the raw connection for #57

gvbgduh:


gvbgduh:
or it might wait before the _support for explicit connection management_ is implemented

tomchristie:
Thanks for this! I reckon we'll want to think through layout a bit more before we get this in -we'd probably want to move beyond a single README.

tomchristie:
I'm going to close this off for now, pending a more comprehensive docs refresh at some point.

PR #67, opened 2019-03-12
Support for explicit connection management #9

gvbgduh:
Support for this:

```connection = await database.connection() ... connection.release()```

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,27 @@ await database.disconnect()
 Connections are managed as task-local state, with driver implementations
 transparently using connection pooling behind the scenes.
 
+## Connections
+
+Connections are managed by async context blocks:
+
+```python
+async with database.connection():
+    ...
+```
+
+For a lower-level transaction API:
+
+```python
+connection = await database.connection()
+try:
+    ...
+except:
+    ...
+finally:
+    connection.release()
+```
+
 ## Transactions
 
 Transactions are managed by async context blocks:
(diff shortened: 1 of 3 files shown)


gvbgduh:
Finally cleaned some mess with repos, hope it looks good, the unittest might look a bit heavy though

tomchristie:
Neato! A couple of thoughts:

* It'd be grand if we could have tests against the public API only, rather than testing internal bits of state. (Haven't exactly thought about how we'd want to do that, I guess it's possible we might want to expose the connected/released state somehow.)
* I'd probably switch it so that `__aexit__` calls `release()` instead of the other way around. That'd mirror transactions then.
* We could *also* support the decorator usage style that we have for transactions. (Although I'd also be okay with that coming as a later follow-up)

gvbgduh:
Thanks!

1. Yeah, there's something wrong using private API in unites I'll adjust this. On the other hand it might be a good indicator of some missing bits in the public one.
2. It definitely would make more sense.
3. Good idea, I'll see how it goes.

gvbgduh:
There're fixes the first 2 points. The unittest looks a bit better.
I'm happy with any case for the connection decorator be it continuation of this one or a separate one.

tomchristie:
Closing off a bunch of PRs as stale.
If that's not the case, then you're super welcome to resubmit this, and we can go from there.

PR #68, opened 2019-03-13
Added support for raw queries using sqlalchemy.text()

pvanliefland:
See #9 

Still pretty basic, not sure if this is the way to go...

### DIFF
--- a/databases/backends/mysql.py
+++ b/databases/backends/mysql.py
@@ -112,10 +112,8 @@ class MySQLConnection(ConnectionBackend):
         finally:
             await cursor.close()
 
-    async def execute(self, query: ClauseElement, values: dict = None) -> typing.Any:
+    async def execute(self, query: ClauseElement) -> typing.Any:
         assert self._connection is not None, "Connection is not acquired"
-        if values is not None:
-            query = query.values(values)
         query, args, context = self._compile(query)
         cursor = await self._connection.cursor()
         try:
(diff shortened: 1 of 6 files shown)


tomchristie:
This is looking *really great*!
Very nearly there.

I think we'll want to push these through to the `Connection` interface instead, so that...

* The `Connection` interface continues to mirror the `Database` interface.
* The query building takes place within the connection.

Technically the `connection` isn't currently public interface, but it *will* become so. (See https://github.com/encode/databases/pull/67) Most users won't need it, but we'll allow explicitly grabbing-and-holding a connection if that's what you want to do, in addition to the current style where connections are obtained and released from the pool as needed, and only hang on forcibly if you're within a transaction.

pvanliefland:
Hi! Something like this?

tomchristie:
Looks great! Are query arguments always the `:query_arg` style with this?

(Don't mind either ways what it uses, just thinking about merging and adding some docs for it. Need to make sure I'm saying the right thing.)

pvanliefland:
Yup, just checked the SA docs, I confirm query arguments follow the `:query_arg` style.

I was thinking of adding something like this to the docs, but not sure where:

## Raw queries

In addition to SQLAlchemy core queries, you can also perform raw SQL queries:

```python
# Execute
query = "INSERT INTO notes(text, completed) VALUES (:text, :completed)"
await database.execute(query, values)

# Execute many
query = "INSERT INTO notes(text, completed) VALUES (:text, :completed)"
values = [
    {"text": "example2", "completed": False},
    {"text": "example3", "completed": True},
]
await database.execute_many(query, values)

# Fetch multiple rows
query = "SELECT * FROM notes WHERE completed = :completed"
rows = await database.fetch_all(query=query, values={"completed": True})

# Fetch single row
query = "SELECT * FROM notes WHERE id = :id"
result = await database.fetch_one(query=query, values={"id": 1})
```

tomchristie:
Let's do it!
Great work 👏🎉

tomchristie:
Version 0.2.0 released 👍

PR #70, opened 2019-03-17
Connection decorator draft

gvbgduh:


gvbgduh:
Connection decorator brings some complexity or might be just redundant.
The matter is that it seems reasonable to use `@database.connection` outside of `async with Database(database_url) as database:` (or `@app.on_event("startup")`) that takes care of connecting to the db and acquiring the connection. So the connection decorator seems to have not a lot of sense here.

For cases where `database.connect()` isn't called implicitly (like in `database = Database(database_url)` the connection decorator should take care of the connecting to the db.

There's a one problem appeared here that the connection_counter might be misled by the call `database.<method: fetch | execute | ...>` if it's accidentally called before.

tests/test_databases.py
```diff
@@ -643,3 +643,36 @@ def run_sync(*args, **kwargs):
...
+
+    async def insert_data_without_decorator():
+        query = notes.insert().values(text="example", completed=True)
+        await database.execute(query)
+
+    # The sqlite backend has no the pool implemented yet
+    # So connection will exist anyway
+    if not str(database_url).startswith("sqlite"):
+        with pytest.raises(AssertionError) as excinfo:
+            await insert_data_without_decorator()
```
gvbgduh: this call will set the connection counter to 1 regardless of the exception.

databases/core.py
```diff
@@ -156,6 +160,21 @@ def __init__(self, backend: DatabaseBackend) -> None:
...
+        @functools.wraps(func)
+        async def wrapper(*args: typing.Any, **kwargs: typing.Any) -> typing.Any:
+            connected_by_decorator = False
+            if not self._backend._pool:
+                await self._backend.connect()
+                connected_by_decorator = True
+
+            async with self:
+                return await func(*args, **kwargs)
+        return wrapper
```
gvbgduh: Having this logic within the decorator isn't very nice, plus there's some ambiguity where the db should be disconnected back, which makes it quite expensive operation though.

databases/core.py
```diff
@@ -141,7 +141,11 @@ def __init__(self, backend: DatabaseBackend) -> None:
         async with self._connection_lock:
             self._connection_counter += 1
             if self._connection_counter == 1:
-                await self._connection.acquire()
+                try:
+                    await self._connection.acquire()
+                except Exception as err:
+                    self._connection_counter -= 1
+                    raise err
```
gvbgduh: this might apply to release as well.

tomchristie:
Closing as stale.

PR #77, opened 2019-04-05
List Tornado as an example async web framework

mivade:


tomchristie:
Yes good call - I wasn't sure if Tornado supported async/await all the way through, since it preceeds 3.5, but looks like it'd work just grand with `databases` as far as I can see.

Eg. https://www.tornadoweb.org/en/stable/guide/structure.html#asynchronous-handlers

PR #79, opened 2019-04-08
Add tests showing error on bulk update with non-raw query

reillysiemens:
I've been battling with an issue today where I was unable to do a bulk update on a table using the non-raw query builder syntax. These tests should showcase that the issue has something to do with how values are handled in `databases.core.Database._build_query`. In particular it seems to be impossible to use `bindparam`s in where clauses with the current functionality. The assertions in `test_queries` fail, but the assertions in `test_raw_queries` pass even though they ought to map to the same SQL.

### DIFF
--- a/tests/test_databases.py
+++ b/tests/test_databases.py
@@ -155,6 +155,26 @@ async def test_queries(database_url):
             assert iterate_results[2]["text"] == "example3"
             assert iterate_results[2]["completed"] == True
 
+            # update
+            query = (notes.update()
+                     .values(completed=sqlalchemy.sql.bindparam('completed'))
+                     .where(notes.c.text == sqlalchemy.sql.bindparam('_text')))
+            values = [
+                {'_text': 'example1', 'completed': False},
+                {'_text': 'example2', 'completed': True},
+            ]
+            results = await database.execute_many(query, values)
+
+            query = notes.select().order_by(notes.c.id.asc())
+            results = await database.fetch_all(query=query)
+            assert len(results) == 3
+            assert results[0]["text"] == "example1"
+            assert results[0]["completed"] == False
+            assert results[1]["text"] == "example2"
+            assert results[1]["completed"] == True
+            assert results[2]["text"] == "example3"
+            assert results[2]["completed"] == True
+
 
 @pytest.mark.parametrize("database_url", DATABASE_URLS)
 @async_adapter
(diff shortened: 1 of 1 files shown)


tomchristie:
Great work on sqauring this away into a pair of test cases.

Does bindparam work correctly for us with non-bulk queries?

reillysiemens:
@tomchristie: If by
> Does bindparam work correctly for us with non-bulk queries?

you mean does something like
```python
query = (notes.update()
         .values(completed=sqlalchemy.sql.bindparam('completed')
         .where(notes.c.text == sqlalchemy.sql.bindparam('_text')))
values = {'_text': 'example1', 'completed': False}
await database.execute(query, values=values)
```
then no, I don't think `bindparam` works correctly for non-bulk queries either.

I think there's an assumption in `databases.core.Database._build_query` that all `values` directly correspond to a table column name. In this case that isn't true since AFAIK SQLAlchemy reserves `bindparam` usage on columns for `SET` and `VALUES` clauses and usage in side a `WHERE` clause requires a prefix like `_` (or something to that effect).

I think `_build_query` tacks on a `.values()` call to queries that already have a `.values()` and that causes a SQLAlchemy error like
```python
sqlalchemy.exc.CompileError: Unconsumed column names: _text
```
as, indeed, `_text` is not a column name on the `notes` table.

pvanliefland:
I just took a look at this to see if it was an easy fix, and I don't think it is... Looking at the [SQLAlchemy documentation for bindparam](https://docs.sqlalchemy.org/en/latest/core/sqlelement.html#sqlalchemy.sql.expression.bindparam), it seems that it is meant to be used with `connection.execute`... As encode/databases does not use SQLAlchemy connections, we won't benefit from the bound parameters handling that happening in SQLALchemy/DBAPI

Additional info here: https://docs.sqlalchemy.org/en/latest/faq/sqlexpressions.html#how-do-i-render-sql-expressions-as-strings-possibly-with-bound-parameters-inlined

tomchristie:
Is #83 the same as this?
Do we just need a bit of change in `_build_query` in order to support `bindparam`?

pvanliefland:
I think it's kind of a won't fix (or won't fix easily)... using `.values()` when building the SQLAlchemy core queries works fine for simple insert/update use cases, but not for the example here or the delete example in #83.

I think the proper solution would be to stop calling `.values()` in `_build_query`, and re-implement the parameters handling magic that happens when you call `connection.execute()` in SQLAlchemy. I took a look at it but cannot really figure out how it is done.

Here is how the example of this PR would be implemented in SQLAlchemy:

```python
query = (notes.update()
                     .values(completed=sqlalchemy.sql.bindparam('completed'))
                     .where(notes.c.text == sqlalchemy.sql.bindparam('_text')))

connection.execute(query, [
                {'_text': 'example1', 'completed': False},
                {'_text': 'example2', 'completed': True},
            ])
```

tomchristie:
Closing this off for now given @pvanliefland's comment.
Let's talk this over on https://gitter.im/encode/community if there's still work that's worth pursuing here.

librehat:
it would be really nice to have the support for `bindparam`. I just ran into the same issue and it doesn't work on either single `execute` nor `execute_many` :(

Issue #78, opened 2019-04-07
I was not able to set min_size or max_size within Postgres


mths0x5f:
I'm trying to set these parameters on last version available on pip, with no success:

```python
    await asyncio.wait_for(connected, loop=loop, timeout=timeout)
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/tasks.py", line 412, in wait_for
    return fut.result()
asyncpg.exceptions.UndefinedObjectError: unrecognized configuration parameter "max_size"
ERROR: Application startup failed. Exiting.
```

tomchristie:
Could you show me more precisely how to replicate the issue, eg...

1. The minimal possible source code that you can use to replicate?
2. The full traceback that is raised?

Will have a look at reopening the issue after that. Thanks!

mths0x5f:
Sorry for the short issue report. I was kinda of sleepy. Now I'm just in doubt if missing something really obvious

I'm running Python 3.7.0 on macOS 10.13.6, installed with Homebrew I think

`app.settings` 
```python
DATABASE_URI = 'postgresql://postgres:postgres@localhost:5432/teste'
DATABASE_MIN_POOL_SIZE = 10
DATABASE_MAX_POOL_SIZE = 50
```

`app.main`
```python
import databases
from fastapi import FastAPI
from app.settings import DATABASE_URI, DATABASE_MIN_POOL_SIZE, DATABASE_MAX_POOL_SIZE

app = FastAPI()

db = databases.Database(f'{DATABASE_URI}?min_size={DATABASE_MIN_POOL_SIZE}&max_size={DATABASE_MAX_POOL_SIZE}')


@app.on_event('startup')
async def startup():
    await db.connect()


@app.on_event('shutdown')
async def shutdown():
    await db.disconnect()
```

When run with `uvicorn app.main:app`, this is the full traceback I get:
```python
INFO: Started server process [16477]
INFO: Waiting for application startup.
ERROR: Exception in 'lifespan' protocol
Traceback (most recent call last):
  File "/Users/matheus/PycharmProjects/teste/venv/lib/python3.7/site-packages/uvicorn/lifespan/on.py", line 43, in main
    await app_instance(self.receive, self.send)
  File "/Users/matheus/PycharmProjects/teste/venv/lib/python3.7/site-packages/starlette/routing.py", line 478, in asgi
    await self.startup()
[...]

PR #71, opened 2019-03-19
Support for additional connection params

gvbgduh:
Refs #69

Issue #69, opened 2019-03-15
Support for connection params in `Database` in addition to the database url


gvbgduh:
There's a small proposal/question about the details for contenting the DB.
There might be a situation where the host might have a complex address, particularly GCP has something like: 
`/<cloudsql>/<project>:<region>:<db_instance>`
Generally, it was solved being passed as a query param `unix_socket=<host>/.s.PGSQL.5432` to the sqlalchemy engine.
It doesn't work with `asyncpg` as it tries to parse the db url and the `:` char violates the parsing logic (when such host is part of the url or passed as environment variable) or the host is empty str when it's passed as `unix_socket` param.

But `asyncpg` can take connections params separately https://magicstack.github.io/asyncpg/current/api/index.html#connection. Params have higher precedence over the url parsing, so they can take place.

I haven't come up with a great idea yet, but it might be done as 

```
class PostgresBackend | DatabaseBackend:
    def __init__(self, database_url: typing.Union[DatabaseURL, str], params: typing.Dict) -> None:
        self._database_url = DatabaseURL(database_url)
        ...
        self._params = params

    async def connect(self) -> None:
        assert self._pool is None, "DatabaseBackend is already running"
        kwargs = self._get_connection_kwargs()
        kwargs.update(self._params)
        self._pool = await asyncpg.create_pool(str(self._database_url), **kwargs)
```

Would it be possible to consider?

Happy to create a PR with a reasonable solution.

gvbgduh:
To be honest, haven't investigated this situation in `aiomysql`.

gvbgduh:
Yeah, I think I've come up with the solution that takes minimum amount of blood.
`DatabaseURL` already does parsing and provides `DatabaseURL.options` kwargs that are passed further. The only problem now that the set of those options is just `{'min_size', 'max_size', 'ssl'}`, so other options are ignored.

So the url `'postgresql://USER:PWD@/DB?host=HOST:project:region:db&port=5432'` should do it, where in `postgres` it can be just extended to recognise other options
`mysql` might have additional backup here, like

```
self._pool = await aiomysql.create_pool(
    host= kwargs.pop("host", None) or self._database_url.hostname,
    ...
    autocommit=kwargs.pop("autocommit", None) or True,
    **kwargs,
)
```

This gives more flexibility and control over it and extends the amount of supported urls.

While I think it might be not a good idea to allow arbitrary key in `kwargs`, it's worth considering to add all possible options per back end.

Any thoughts/objections?

[...]

PR #85, opened 2019-04-17
Connection options

### LINKED ISSUE #84: How to pass an SSL context to the postgresql backend?
Hi,

It appears that databases only accepts a boolean to indicate if you want to connect over TLS to postgresql. But, the [asyncpg driver supports to pass your own context](https://magicstack.github.io/asyncpg/current/api/index.html#connection), which is necessary when you have your own certs.

This make integrating in some environment a showstopper (our environment relies on TLS internally).

Was there a specific issue for not allowing passing a context?

Cheers,
- Sylvain

tomchristie:
Allows `database = Database(url, **connection_options)`

Refs #78 (tho it doesn't strictly close it, since it's a fix ,but passing the options in the URL will still break.).
Refs #71, #69 (@gvbgduh is there anything you think needs to follow on from this? Do we need to drop options out of the URL string or something along those lines?)
Closes #84.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ It allows you to make queries using the powerful [SQLAlchemy Core][sqlalchemy-co
 expression language, and provides support for PostgreSQL, MySQL, and SQLite.
 
 Databases is suitable for integrating against any async Web framework, such as [Starlette][starlette],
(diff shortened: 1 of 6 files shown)


tomchristie:
Version 0.2.2

PR #87, opened 2019-04-17
Add PgSQL record access by attribute

Lawouach:
It's fairly usual to access a record's content vy its attributed: `myrecord.id`. While this works with the SQLite backend, it's not possible now with the PostgreSQL one. This fix replicates what SA does on a RowProxy.

Signed-off-by: Sylvain Hellegouarch <sh@defuze.org>

### DIFF
--- a/databases/backends/postgres.py
+++ b/databases/backends/postgres.py
@@ -119,6 +119,12 @@ class Record(Mapping):
     def __len__(self) -> int:
         return len(self._column_map)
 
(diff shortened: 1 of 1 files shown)


tomchristie:
Going to pass on this. Okay for an ORM-level construct, but it's often awkward since it can clash with built-in attributes.

Lawouach:
That's fair enough indeed.

PR #95, opened 2019-05-01
Support absolute paths for SQLite

blueyed:
This makes it compatible with dj-database-url [1].

1: https://github.com/kennethreitz/dj-database-url#id6

### DIFF
--- a/databases/core.py
+++ b/databases/core.py
@@ -351,7 +351,10 @@ class DatabaseURL:
 
     @property
     def database(self) -> str:
-        return self.components.path.lstrip("/")
+        path = self.components.path
+        if path[:1] == "/":
+            return path[1:]
+        return path
 
     @property
     def options(self) -> dict:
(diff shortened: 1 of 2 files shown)


databases/core.py
```diff
@@ -351,7 +351,10 @@ def netloc(self) -> typing.Optional[str]:
 
     @property
     def database(self) -> str:
-        return self.components.path.lstrip("/")
+        path = self.components.path
+        if path[:1] == "/":
+            return path[1:]
+        return path
```
blueyed: This is not covered.  Let me know if you want to take this in the first place.

OTOH I am not using SQLite anymore, so feel free to just close it also.. ;)

tomchristie: I reckon either you or I oughta follow up on it given it's now also being reported elsewhere, yup. #115

Happy to pick it up if you've moved on for it.

tomchristie:
Superseeded by #122

Thanks so much for this!

PR #97, opened 2019-05-07
postgres: use wait_for in disconnect

blueyed:
This is recommended by asyncpg [1].

1: https://github.com/MagicStack/asyncpg/blob/92c2d81256a1efd8cab12c0118d74ccd1c18131b/asyncpg/pool.py#L655-L656


blueyed:
Note that there is also a timeout for `acquire`, which gets used by `release` then also: https://github.com/MagicStack/asyncpg/blob/92c2d81256a1efd8cab12c0118d74ccd1c18131b/asyncpg/pool.py#L560-L564

tomchristie:
This is a good call yup.
Presumably we also want to add the timeout to `acquire` right?
Any thoughts about adding `timeout` to the API someplace, and using that value?

databases/backends/postgres.py
```diff
@@ -66,7 +67,7 @@ async def connect(self) -> None:
 
     async def disconnect(self) -> None:
         assert self._pool is not None, "DatabaseBackend is not running"
-        await self._pool.close()
+        await asyncio.wait_for(self._pool.close(), timeout=30)
```
vmarkovtsev, line 70: As @tomchristie wrote, it is a good idea to make the timeout configurable, otherwise, 30 seconds can be too much for everybody.

blueyed:
@vmarkovtsev 

> I can take care of finishing this PR if you don't have time right now. WDYT?

Sure, please go ahead.
(Sorry for the late reply, of course)

taybin:
Any progress on this?

PR #100, opened 2019-05-14
add py.typed to let mypy notice inline types

shamrin:
Without this file I get mypy error at `import databases`:

```
my_code/db.py:5: error: Cannot find module named 'databases'
my_code/db.py:5: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
```

The error went away after I added empty `py.typed` into `databases` folder:

```
touch /Users/.../virtualenvs/backend-JkxMzBya/lib/python3.7/site-packages/databases/py.typed
```

(I've installed `databases` via `pipenv install databases`.)


tomchristie:
Thanks!
We probably also need `zip_safe=False` in the setup.py.
See https://github.com/encode/starlette/pull/455

shamrin:
> We probably also need `zip_safe=False` in the setup.py.

Done!

tomchristie:
👍

Issue #102, opened 2019-05-24
Result iteration on postgres requires a transaction


sm-Fifteen:
I'm not sure if this is meant to be the intended behavior or not, since iterate() isn't very documented as of right now, but attempting to use it without creating a transaction first causes asyncpg to throw the following: `asyncpg.exceptions.NoActiveSQLTransactionError: cursor cannot be created outside of a transaction`.

The issue can easily be reproduced with the following snippet. I haven't tested other backends to see if the same thing occurs with SQLite and MySQL.

```py
from databases import Database
import asyncio

mydb = Database('postgresql://localhost:5433/testdb11')

async def test():
    await mydb.connect()

    async for row in mydb.iterate(query='SELECT * FROM (VALUES (1), (2), (3), (4), (5)) as t(num)'):
        print(row)

    await mydb.disconnect()

loop = asyncio.get_event_loop()  
loop.run_until_complete(test())  
loop.close()
```

SQLAlchemy doesn't require the use of transactions to iteratively load the results of a query, and the documentation doesn't specify this constraint, so transactions probably shouldn't be enforced here either as far as the public API for functions goes.

tomchristie:
Okay. I guess we could implicitly create a transaction within the postgres driver code if there isn't already one exsiting.

tomchristie:
Best thing to help progress this issue would be a pull request with a failing test case.
That'd also help us identify if we have the same issue against other drivers or not.

tomchristie:
Closed via #111 - We now just always wrap iteations up inside a transaction, whatever. (Which may be a sub-transaction/savepoint) Seems pretty reasonable.

PR #109, opened 2019-06-22
Test case for iterating outside a transaction for #102

sm-Fifteen:
The asyncpg driver relies on server-side cursors without hold for iteration, so they don't work unless the query is wrapped within a transaction. This is mentionned in both [their documentation][1] and their [test suite][2].

This seems to be the only driver with that quirk, so this test case ensures that databases uniformises that behavior across all supported engines.

[1]: https://magicstack.github.io/asyncpg/current/api/index.html#cursors
[2]: https://github.com/MagicStack/asyncpg/blob/2d76f50dccf35cf2f1d70b41ebd6198d2dfff8d7/tests/test_cursor.py#L31-L49

### DIFF
--- a/tests/test_databases.py
+++ b/tests/test_databases.py
@@ -699,3 +699,24 @@ async def test_database_url_interface(database_url):
     async with Database(database_url) as database:
         assert isinstance(database.url, DatabaseURL)
         assert database.url == database_url
+
+
+@pytest.mark.parametrize("database_url", DATABASE_URLS)
+@async_adapter
+async def test_iterate_outside_transaction(database_url):
+    """
+    Ensure `iterate()` works even without a transaction on all drivers.
+
+    The asyncpg driver relies on server-side cursors without hold
+    for iteration, which requires a transaction to be created.
+    This is mentionned in both their documentation and their test suite.
+    """
+    async with Database(database_url) as database:
+        query = "SELECT * FROM (VALUES (1), (2), (3), (4), (5)) as t"
+        iterate_results = []
+
+        async for result in database.iterate(query=query):
+            iterate_results.append(result)
+
+        assert len(iterate_results) == 5
+        assert iterate_results == [(1,), (2,), (3,), (4,), (5,)]
--- a/tests/test_databases.py
+++ b/tests/test_databases.py
@@ -703,16 +703,49 @@ async def test_database_url_interface(database_url):
 
 @pytest.mark.parametrize("database_url", DATABASE_URLS)
 @async_adapter
-async def test_iterate_outside_transaction(database_url):
+async def test_iterate_outside_transaction_with_values(database_url):
     """
     Ensure `iterate()` works even without a transaction on all drivers.
-
     The asyncpg driver relies on server-side cursors without hold
     for iteration, which requires a transaction to be created.
     This is mentionned in both their documentation and their test suite.
     """
+
+    database_url = DatabaseURL(database_url)
+    if database_url.dialect == "mysql":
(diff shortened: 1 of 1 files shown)


sm-Fifteen:
Looks like I forgot to check whether or not MySQL supported `SELECT * FROM (VALUES ...)`. I guess SQL F641 isn't universally supported yet.

sm-Fifteen:
So while this test works a expected on the Postgres backend, I can't seem to get it to work properly on the SQLite backend.

```py
@async_adapter
async def test_iterate_outside_transaction(database_url):
        query = "CREATE TEMPORARY TABLE no_transac(num INTEGER)"
        await database.execute(query)

        query = "INSERT INTO no_transac(num) VALUES (1), (2), (3), (4), (5)"
        await database.execute(query)

        query = "SELECT * FROM no_transac"
        iterate_results = []

        async for result in database.iterate(query=query):
            iterate_results.append(result)

        assert len(iterate_results) == 5
        assert iterate_results == [(1,), (2,), (3,), (4,), (5,)]
```

It dies at `INSERT INTO no_transac(num) VALUES (1), (2), (3), (4), (5)` saying `no_transac` does not exist. Running an equivalent script on SQLite directly produces the expected result, however.

Weirder still is that this works:
```py
import aiosqlite

@async_adapter
async def test_sqlite_temp_table():
    async with aiosqlite.connect('test_db.sqlite') as db:
        await db.execute("CREATE TEMPORARY TABLE no_transac(num INTEGER)")
        await db.execute('INSERT INTO no_transac(num) VALUES (1), (2), (3), (4), (5)')

        iterate_results = []

        async with db.execute('SELECT * FROM no_transac') as cursor:
            async for row in cursor:
                iterate_results.append(row)
        
        assert len(iterate_results) == 5
        assert iterate_results == [(1,), (2,), (3,), (4,), (5,)]
```

@tomchristie: Bug with the sqlite backend?

tomchristie:
Don't know. We could just skip this test for sqlite for now.

Ideally it'd be good if we could have proper markers for test cases where we want to only run against particular driver implementations.

sm-Fifteen:
I added both versions of the test, skipping one or the other based on whether the dialect is detected as SQLite or MySQL. Postgres works on both versions, and the test fails as expected. This confirms that the issue only occurs on asyncpg.

Should the SQLite bug regarding temporary tables be raised as a separate issue, considering SQLAlchemy doesn't have any interfaces for these?

tomchristie:
> Should the SQLite bug regarding temporary tables be raised as a separate issue

No - I don't believe that's really an issue right? Not all database implementations will support all SQL syntaxes - that's nothing to do with this project.

> I added both versions of the test, ... This confirms that the issue only occurs on asyncpg.

Great work!

So: Resolution will be that the postgres driver needs to check if it's in a transaction or not, and temporarily open one if not.

tomchristie:
Following up on this now in #111 - thanks!

sm-Fifteen:
> > Should the SQLite bug regarding temporary tables be raised as a separate issue
> 
> No - I don't believe that's really an issue right? Not all database implementations will support all SQL syntaxes - that's nothing to do with this project.

That's the thing, when I tried using `aiosqlite` directly to run the same sequence on commands (see above), the test did pass, so it would appear to be a problem with the backend for aiosql that's causing this. I haven't investigated much beyond that, though, so I don't know if there are any other cases that would be affected.

tomchristie:
@sm-Fifteen Ah, curious. Yeah welcome to open a separate issue for that, ideally with an example of "this is what works when using the driver directly" vs. "this is what doesn't work when using `databases`".

PR #111, opened 2019-06-25
Iterating outside of transaction

tomchristie:
Following up #109


tomchristie:
We now always iterate within a transaction.
I think this is probably a perfectly reasonable policy for us to have here.

PR #112, opened 2019-06-25
use native decimal (numeric) type

### LINKED ISSUE #113: Native decimal support
Hi,
first thank you for this great library. I was missing native decimal support (decimals are rounded) and sketched a solution in https://github.com/encode/databases/pull/112 - seems like it can be safely enabled for Postgres and MySQL. Hope this would be useful for you.
Cheers
Jakub

coobas:
Enables Postgresql and MySQL native support for Decimal (NUMERIC) type, which improves precision for decimals.

Closes #113


tomchristie:
This is great stuff. Yes please!

Issue #117, opened 2019-07-01
performance issue


euri10:
I have a big performance issue, 1.5s using the raw query, more than 16s using the sqlalchemy core version of it.

```
    raw_query = str(stmt.compile(dialect=postgresql.dialect(),  compile_kwargs={"literal_binds": True}))

    conn = await asyncpg.connect(str(databaselake.url))
    direct0 = time.time()
    direct_result = await conn.fetch(raw_query)

    await conn.close()
    direct1 = time.time()
    print(f"direct: {direct1-direct0}")

    raw0 = time.time()
    raw_result = await databaselake.fetch_all(raw_query)
    raw1 = time.time()
    print(f"raw: {raw1-raw0}")

    stmt0 = time.time()
    result = await databaselake.fetch_all(stmt)
    stmt1 = time.time()
    print(f"stmt: {stmt1 - stmt0}")
    print("result ok")
```

direct: 1.3257312774658203
raw: 1.4148566722869873
stmt1: 19.825509071350098

`direct` and `raw` both use `raw_query` which is the fully compiled version of `stmt` 

They have similar performance, in line with what I get from entering the query in psql, around 1.5s

`direct` uses asyncpg directly, this was just to test, while `raw` uses databases

now the fetch_all version using the sqlachemy core `stmt` takes 16 times more time...

the query returns about 21 columns and 4500 rows. it's about 80 000 characters long, not sure it makes sense to paste it here but maybe there's something else I havn't considered

edit: the raw statement looks like this

```
select l.field1, l.field2, l.field3, ,......,  l.field21
from lake l
         join (
    select x.unnest, x.ordinality
    from unnest(array ['ticker1','ticker2', ....., 'ticker4299', 'ticker4300']) with ordinality
             as x (unnest, ordinality)) as r on l.ticker = r.unnest
where time = '2019-06-20'
order by r.ordinality;
```

euri10:
even this one takes 1.9s
```
    engine = create_engine(str(databaselake.url))
    c = engine.connect()
    psycopg0 = time.time()
    psycopg_result = c.execute(stmt).fetchall()
    psycopg1 = time.time()
    print(f"psycopg: {psycopg1-psycopg0}")
    c.close()
```

tomchristie:
I guess the first thing to do here would be to try to get some finer grain information about:

* Where the extra time is going there?
* If the auto-compiled statement really is exactly the same as the one you're manually compiling or not?
* How the compilation differs in the two cases (If it does).

[...]

PR #120, opened 2019-07-03
Slow

euri10:
PR for discussion on #117


tomchristie:
I really can't help out with the issue presented in it's current state.

Very happy to try to get involved, but you'd need to work on first reducing this right down to the absolute simplest possible reproducible case, so that it's been properly isolated into a clear issue.

Right now I've got no way of knowing if this is some issue with the way you're using SQLAlchemy core, or if it's actually an issue in something to do with `databases` itself.

PR #121, opened 2019-07-08
Fix record iteration

coobas:
Hi again,
I ran into an issue with row columns (`row.keys()` would be empty) from a raw (text) query ('SELECT * FROM ...'). It seems to me it can be fixed by this MR - please have a look. As a side effect I had to adapt some tests for pytest 5, it works with 4.6 as well though.
Cheers
Jakub

### DIFF
--- a/databases/backends/postgres.py
+++ b/databases/backends/postgres.py
@@ -115,10 +115,10 @@ class Record(Mapping):
         return raw
 
     def __iter__(self) -> typing.Iterator:
-        return iter(self._column_map)
+        return iter(self._row.keys())
 
     def __len__(self) -> int:
-        return len(self._column_map)
+        return len(self._row)
 
 
 class PostgresConnection(ConnectionBackend):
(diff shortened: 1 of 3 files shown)


tomchristie:
Looks decent. Would you also be able to include a test case that fails without this change, but passes with it?

coobas:
Of course, a test should have been there! I've added one - and checked it fails when without the fix (https://travis-ci.org/coobas/databases/builds/556169826)

tomchristie:
Looks great! I think the docstring needs a tweak right? I don't think it's quite reflective of what the test is actually for. After that I think we're good to go.

tomchristie:
Alrighty, I've tweaked the docstring. Once Travis completes lets get this in!
Nice one.

coobas:
Thanks, and sorry for the docstring. I just started looking into that + resolving conflicts but you've been faster :)

tomchristie:
Version 0.2.5 now available on PyPI.
Great work!

coobas:
Thanks!

PR #124, opened 2019-07-10
Aiopg

gvbgduh:


gvbgduh:
Hi @tomchristie, I was finally able to finish the first turn around and there're some issues appeared.

1. `postgresql+aiopg` doesn't work for sqlalchemy as it doesn't recognise that driver (yet?) https://docs.sqlalchemy.org/en/13/core/engines.html#postgresql
so, I took the `postgresql+psycopg2` one for start.
I have in mind a couple more ideas how to approach:
- transform `+aiopg` to `+psycopg` in the `__repr` or some cast, but it doesn't feel right
- pass/specify a particular backed to the `Database.__init__`, it might allow more flexibility later, but it's not a small architectural change and I'm not fully sure how it's aligned.

Also there might be some confusion as `asyncpg` is implicit choice for Postgres (even though I believe a good one :)).

2. A bit different handling for JSON fields values, but I need to cripple check.

3. `lastrowid` seems to be not exactly the record id, but a local op id - https://aiopg.readthedocs.io/en/stable/core.html#aiopg.Cursor.lastrowid. But I'd rather triple check it as well.

4. A bit different approach for transactions, basically only raw queries can allow this.

In overall, it's pretty close to `mysql` backend, but there're some differences in details.
I also spotted a couple of places for optional refactoring  - mostly moving some stuff to conftest for unittests, but there're a couple of methods that are some across all backend implementations.

Yeah, it would great to have any feedback, recommendations, complains.

tomchristie:
I suppose sqlalchemy should only see it as `postgresql`. We're not using their drivers, so that's the best information we can give it.

gvbgduh:
Yes, I see, it clarifies and makes it easier.
Ok, now I would really appreciate a review if it could be possible.

mpyatishev:
@tomchristie, @gvbgduh when do you plan to merge this PR?

eien:
Yup, it'd be great to get this in!

tests/test_connection_options.py
```diff
@@ -52,3 +53,29 @@ def test_mysql_explicit_ssl():
...
+def test_aiopg_explicit_pool_size():
+    backend = AiopgBackend(
+        "postgres+aiopg://localhost/database", min_size=1, max_size=20
+    )
+    kwargs = backend._get_connection_kwargs()
+    assert kwargs == {"minsize": 1, "maxsize": 20}
+
+
+def test_aiopg_ssl():
+    backend = AiopgBackend("postgres+aiopg://localhost/database?ssl=true")
```
tomchristie: Looks like we're expecting `postgres+aiopg` in some places and `postgresql+aiopg` in others.
I'd be okay with us accepting either or both.

gvbgduh: Right, yeah, it's a typo, but we only expect `postgresql+aiopg` as of now, at least there's only this mapping. it's not triggered in these tests as don't make any use of it. I'll update urls to avoid confusion.

tests/test_databases.py
```diff
@@ -301,10 +306,17 @@ def run_sync(*args, **kwargs):
...
             assert isinstance(pk, int)
 
-            query = notes.select().where(notes.c.id == pk)
-            result = await database.fetch_one(query)
-            assert result["text"] == "example1"
-            assert result["completed"] == True
+            # Apparently for `aiopg` it's OID that will always 0 in this case
+            # As it's only one action within this cursor life cycle
+            # It's recommended to use the `RETURNING` clause
+            # For obtaining the record id
```
tomchristie: This is a good example of where we ought to be returning a more structured type from `.execute` - that way we could have different drivers either support or not support `result.row_id`.

tests/test_databases.py
```diff
@@ -503,8 +515,13 @@ def run_sync(*args, **kwargs):
         async with database.transaction(force_rollback=True):
             # execute()
             query = session.insert()
-            values = {"data": {"text": "hello", "boolean": True, "int": 1}}
-            await database.execute(query, values)
+            data = {"text": "hello", "boolean": True, "int": 1}
+            values = {"data": data}
+
+            if database.url.scheme == "postgresql+aiopg":
+                await database.execute(query, {"data": json.dumps(data)})
```
tomchristie: Can we handle this automatically for our users?
Would that be a sensible thing to do here?

gvbgduh: Not sure it's possible in all cases, as I would need to know that the column of a type json, so it'll require either a lookup to the sqlalchemy's metadata or to the DB schema. But I'll have a look at how easy it can be.

tests/test_databases.py
```diff
@@ -666,42 +683,47 @@ def run_sync(*args, **kwargs):
...
                 # execute_many()
                 values = [("example2", False), ("example3", True)]
 
-                if str(database_url).startswith("mysql"):
+                if database.url.scheme == "mysql":
                     cursor = await raw_connection.cursor()
                     await cursor.executemany(insert_query, values)
+                elif database.url.scheme == "postgresql+aiopg":
+                    cursor = await raw_connection.cursor()
+                    # No async support for `executemany`
```
tomchristie: Okay, in these kinds of cases we should still support the `.executemany` API, even if all it *actually* does is `for value in values: ...execute...` under the hood.

(We can document limitations of particular drivers, but it'd def. be a benefit if we can keep the API the same, even if there's different performance characteristics in some cases.)

tests/test_integration.py
```diff
@@ -30,6 +30,8 @@ def create_test_database():
         database_url = DatabaseURL(url)
         if database_url.dialect == "mysql":
             url = str(database_url.replace(driver="pymysql"))
+        elif database_url.driver == "aiopg":
```
tomchristie: Let's check the `.scheme` here instead in these conditions.

tests/test_databases.py
tomchristie: Can we change our query here to make it consistently return the pk across all our implementations maybe?

gvbgduh: Yes, sounds reasonable indeed.
There're two ways though as I can see:
- provide pure driver experience with minimal overhead, hence it'll vary from driver to driver, but will keep it as simple as possible if it can matter, or
- provide unified API and implement the missing bits providing the missing functionality, but allocating the additional overhead

tomchristie:
Yes yes yes! Let's make this a priority.
Some review comments above - @gvbgduh feel free to keep nudging me until we've got this one in!

gvbgduh:
Awesome, thanks @tomchristie!

tests/test_databases.py
gvbgduh: Oh, just 25 secs later, ok, let's follow the second path, all in all, there's always a way to jump to the raw driver and get the full manual control over it.

tests/test_databases.py
gvbgduh: Yeah, I see and agree, will try my best.

gvbgduh: Ok, it's the test for the raw driver across different backends, but `aiopg` has the implementation for `execute_many` and yeah it's with the `for` loop.

tests/test_integration.py
gvbgduh: fixed

tests/test_databases.py
gvbgduh: @tomchristie I'm thinking quite hard about it. There's a function in postgres - `lastval()` that can be executed to retrieve the value, but it's not reliable as there're circumstances where it can return the wrong value. The `returning` clause is the best way to do it, but we can't assume how the PK column is called and `returning *` won't be consistent with other back-ends.
The

tests/test_databases.py
tomchristie: > it'll require either a lookup to the sqlalchemy's metadata or to the DB schema

If it means a lookup then let's not do it. I'm just not clear why `JSONField` isn't already handling the serialization for us in this case. Might need some more looking into?

tests/test_databases.py
tomchristie: Grand, let's just hook our backend implementation up to that then, right?

tests/test_databases.py
gvbgduh: yeah, I'm dealing with it now, there're several facets to approach it.
it can work canonically, but the field has to be `sqlalchemy.dialects.postgresql.JSON`, but `aiopg` returns back (over `select`) the decoded value and it breaks the `RowProxy` as it expects string.
so, i'm looking for compromise/settings to get the balance

tests/test_databases.py
gvbgduh: All should be in place! I have a unittest to test it for `database.execute_many`.

gvbgduh: this case is the battle test of the raw driver, so it replicates backend implementations in som places.

tests/test_databases.py
gvbgduh: ok, it was the matter of the dialect class, I reused the one that's in use for the integration with `sa` part of `aoipg` it got back to normal.

gvbgduh:
@tomchristie I think it's in good shape now. There's only one open question - `lastrowid`, but I'm not sure about the universal, consistent and reliable solution for that. Happy to follow up on it in the next PR if it's more reasonable.

gvbgduh:
hi @tomchristie! A gentle nudge here, I think it's mergeable. Happy to follow up on any appeared issues anyway.

StephenCarboni:
What would be the point of this if databases already works with asyncpg?

gvbgduh:
There's a bunch of different reasons, but the main point is `asyncpg` and `aiopg` are totally different, including underlying lower-level drivers, internal and external APIs, and even query syntax details.
`asyncpg` is written from scratch without dependencies, while `aiopg` is based on `psycopg2` which is based on `libpq` and shares the same (or very similar) API with `aiomysql`/`aiosqllite`/`aioodbc`.
Despite personal preferences, it can also cover different migration strategies between DBs, or from sync to async, when the time is the critical factor for example.

Issue #101, opened 2019-05-21
dict(record) in postgres (asyncpg) returns an empty dict


rlittlefield:
my version: 0.2.2
asyncpg version: 0.18.3
functions: database.fetch_one, database.fetch_all

`dict(user_row)` when user_row is a `databases.backends.postgres.Record` object returns an empty dict for me when using postgres through asyncpg.

`dict(user_row.__dict__['_row'])` works, so I know that at least in my case the record has successfully loaded and there is a dict somewhere. dict style `user_row['id']` works just fine.

rlittlefield:
It also seems that user_row will be falsey, such that `bool(user_row)` will be `False`. This makes a common check to see if a row was returned fail.

rlittlefield:
This can be worked around by checking for `if user_row is None`, but when dealing with a row object it is probably expected that any row returned is truthy.

rlittlefield:
https://github.com/encode/databases/blob/c008b2af890d245f35d27d42f5851d329ce2d45a/databases/backends/postgres.py#L96

It looks like this would work if the postgres backend's Record class fully implemented the Mapping interface by adding a `.keys()` method:

```
    def keys(self):
        if not self._column_map:
            return self._row.keys()
        return self._column_map.keys()
```

It appears in local development that adding that method to the Record makes the tests skip over `__getitem__` entirely, which makes me a bit concerned that things like numeric column addresses won't work. Right now you can't use `my_func(**record)` without a `keys()`, because it expands to an empty dict. With the keys above, you could probably still ask for a number column one at a time with square brackets though. We just can't have it expand a column into both a text name and an integer index.

tomchristie:
> things like numeric column addresses won't work

Okay. It's not obvious to me that we ought to be supported that as part of the API in any case. Any reason we shouldn't just be explicit that it returns a mapping interface, and not have any expectation of integer-indexing. Thoughts?

rlittlefield:
If numeric columns are needed as a feature, it might make more sense to expose in a different way, like `record.index(n), or like `record.tuple[1]`. Overloading named/index technically works, but splitting them would be more explicit.

If it really isn't a supported or used feature, it could just be dropped. I always use names because I'm worried about accidentally shifting a column by one.

tomchristie:
> If numeric columns are needed as a feature, it might make more sense to expose in a different way, like record.index(n), or like record.tuple[1]

That's a good call, yup

tomchristie:
Considering this for the API:

[...]

PR #131, opened 2019-07-25
Fix fetch_val for raw queries in Postgres backend

pvanliefland:
As of now, `fetch_val` does not work for raw queries in Postgres. Possibly related to the discussion in #101.

This is a simple fix that will distinguish between `row[0]` and `row['foo']` in the case of raw queries.

```python
await database.fetch_val("SELECT foo FROM bar")
```

in `core.py`:
```python
async def fetch_val(
    self,
    query: typing.Union[ClauseElement, str],
    values: dict = None,
    column: typing.Any = 0,
) -> typing.Any:
    async with self._query_lock:
        row = await self._connection.fetch_one(self._build_query(query, values))
    return None if row is None else row[column]
```

in `backends/postgres.py`:
```python
def __getitem__(self, key: typing.Any) -> typing.Any:
    if len(self._column_map) == 0:  # raw query
        return self._row[tuple(self._row.keys()).index(key)]
```

In this case, key is `0`, while `tuple(self.row.keys())` is `('id',)`, which raises `ValueError: tuple.index(x): x not in tuple`

### DIFF
--- a/databases/backends/postgres.py
+++ b/databases/backends/postgres.py
@@ -96,7 +96,7 @@ class Record(Mapping):
 
     def __getitem__(self, key: typing.Any) -> typing.Any:
         if len(self._column_map) == 0:  # raw query
-            return self._row[tuple(self._row.keys()).index(key)]
+            return self._row[key if type(key) is int else tuple(self._row.keys()).index(key)]
         elif type(key) is Column:
             idx, datatype = self._column_map_full[str(key)]
         elif type(key) is int:
--- a/tests/test_databases.py
+++ b/tests/test_databases.py
@@ -202,6 +202,14 @@ async def test_queries_raw(database_url):
             assert result["text"] == "example2"
             assert result["completed"] == False
 
+            # fetch_val()
+            query = "SELECT completed FROM notes WHERE text = :text"
+            result = await database.fetch_val(query=query, values={"text": "example1"})
+            assert result == True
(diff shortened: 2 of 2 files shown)


gvbgduh:
@pvanliefland `asyncpg` actually has `fetchval` method, in contradiction to `mysql` and `sqlite`, so instead of
```return self._row[key if type(key) is int else tuple(self._row.keys()).index(key)]```
it's possible to implement it
in `postgres.py`
```
    async def fetch_val(self, query: ClauseElement) -> typing.Optional[typing.Mapping]:
        assert self._connection is not None, "Connection is not acquired"
        query, args, result_columns = self._compile(query)
        row = await self._connection.fetchval(query, *args)
        if row is None:
            return None
        return row
```
and in `core.py`
```
    async def fetch_val(
        self,
        query: typing.Union[ClauseElement, str],
        values: dict = None,
        column: typing.Any = 0,
    ) -> typing.Any:
        async with self._query_lock:
            row = await self._connection.fetch_val(self._build_query(query, values))
        return row
```
it also requires implementation of `fetch_val` in other backends in a way it's implemented in `core.py` now.
A bit more work, but it gets rid of this overhead for postgres as `asyncpg` just returns the value.

pvanliefland:
@gvbgduh good idea, will do :)

gvbgduh:
awesome!

blueyed:
For reference: https://github.com/encode/databases/pull/132

PR #132, opened 2019-07-26
Use backend native fetch_val() implementation when available

pvanliefland:
This has the added benefit of fixing `fetch_val` for raw queries in Postgres (which was my initial motivation).


tomchristie:
Cool - can you describe the difference in behavior from a user perpective here?

tests/test_databases.py
```diff
@@ -202,6 +208,16 @@ def run_sync(*args, **kwargs):
...
             assert result["completed"] == False
 
+            # fetch_val()
+            query = "SELECT completed FROM notes WHERE text = :text"
+            result = await database.fetch_val(query=query, values={"text": "example1"})
+            assert result == True
+            query = "SELECT completed FROM notes WHERE text = :text"
+            result = await database.fetch_val(
+                query=query, values={"text": "example1"}, column="completed"
+            )
```
tomchristie: Should this read `query = "SELECT * FROM ..."` and then use `column="completed"`?
It's not obvious at first sight why adding `column=...` makes sense here otherwise?

pvanliefland: You're right, I changed it

pvanliefland:
> Cool - can you describe the difference in behavior from a user perpective here?

Not a lot really, but it fixes `fetch_val` for raw queries (does not work at the moment). See my comment here: https://github.com/encode/databases/issues/101#issuecomment-515097197

pvanliefland:
@tomchristie are you ok with this implementation? I think it's ready to be merged

tomchristie:
Looks good. I think I've successfully resolved a merge conflict, let's see where the tests are at...

tomchristie:
Ace, let's have it!

PR #139, opened 2019-09-11
Fix typo (tests are missing!)

abetkin:


tomchristie:
Heya - thanks for this!
Could we get a test case and a brief explanation?

abetkin:
Hi! A test for this would be a selectable query with parameters, but not a raw sql query. Something like this:

```python
database = Database("postgresql://postgres:postgres@localhost/ponydb")

async def mytest():
    await database.connect()
    p = bindparam('start')
    query = select([mytable]).where(mytable.c.title.startswith(p))
    rows = await database.fetch_all(query=query, values={'start': 'A'})
    for row in rows:
        print(row)
```

abetkin:
@tomchristie ping

garyd203:
Hey @abetkin thanks for fixing this, I ran into the same problem :-). FWIW I have worked around the problem by doing `database.fetch_all(select_query.params(foo=1))`

By the way, I notice that the test run failed simply because of file formatting in `test_databases.py` (perhaps just not enough new lines).

tomchristie:
There's a linter failure on this pull request at the moment.
Running `scripts/lint` would resolve this.

PR #147, opened 2019-10-08
Add postgres dialect

### LINKED ISSUE #146: Support 'postgres' as an alternate dialect name
We may want to support both `postgres://` *and* `postgresql://` as valid dialect names.

Eg.

* Heroku Postgres defaults to using the `postgres` variant.
* [dj-database-url](https://github.com/jacobian/dj-database-url) appears to support both usages.

theruziev:
Fix #146 
Hope, I understood the issue correctly. BTW, this solution works for me


gvbgduh:
@tomchristie should be good to merge

gvbgduh:
thanks @theruziev!

Issue #141, opened 2019-09-16
Interval type column raises TypeError on record access


steinitzu:
Hi

I found an issue when selecting from a table with an [`Interval`](https://docs.sqlalchemy.org/en/13/core/type_basics.html#sqlalchemy.types.Interval) column, this should parse to a `timedelta` object.
When used with databases, sqlalchemy raises a `TypeError` when trying to access the column in the result:

```python
~/projects/dummy/scripts/interval.py in <module>
     33 
     34 loop = asyncio.get_event_loop()
---> 35 result = loop.run_until_complete(do())

/usr/lib/python3.7/asyncio/base_events.py in run_until_complete(self, future)
    577             raise RuntimeError('Event loop stopped before Future completed.')
    578 
--> 579         return future.result()
    580 
    581     def stop(self):

~/projects/dummy/scripts/interval.py in do()
     28 
     29     row = await db.fetch_one(thing.select())
---> 30     result = row["duration"]  # This fails
     31     return result
     32 

~/projects/dummy/.venv/lib/python3.7/site-packages/databases/backends/postgres.py in __getitem__(self, key)
    112 
    113         if processor is not None:
--> 114             return processor(raw)
    115         return raw
    116 

~/projects/dummy/.venv/lib/python3.7/site-packages/sqlalchemy/sql/sqltypes.py in process(value)
   1927                 if value is None:
   1928                     return None
-> 1929                 return value - epoch
   1930 
   1931         return process

TypeError: unsupported operand type(s) for -: 'datetime.timedelta' and 'datetime.datetime'
```

It works fine when using the same table with sqlalchemy engine directly.  
Suspecting it's this line in the result processor that's returning truthy result with databases for some reason, but haven't looked into it thoroughly: https://github.com/zzzeek/sqlalchemy/blob/master/lib/sqlalchemy/sql/sqltypes.py#L1915
My understanding is that the result processor should not be executed for postgres cause it has a native interval type.

Here is a script to reproduce the problem:

```python
import asyncio
from datetime import timedelta

from sqlalchemy import Column, create_engine, MetaData, Table, Interval
from databases import Database

DB_URL = "postgresql://dev:dev@localhost:6808/dev"

metadata = MetaData()

thing = Table(
    "thing",
    metadata,
    Column("duration", Interval, nullable=False),
    extend_existing=True,
)


async def do():
    db = Database(DB_URL)
    if not db.is_connected:
        await db.connect()

    engine = create_engine(DB_URL)
[...]

PR #149, opened 2019-10-11
Fix Interval column

steinitzu:
This is the first step to fix Interval columns ( see #141 )

This fails due to `result_processor` being incorrectly called with postgres.

Next is to figure out what sqlalchemy does differently, and try to replicate that in databases


gvbgduh:
Lovely!

tomchristie:
Really nice to see this as a first step here. I'm assuming I should close this off as stale, but would be fab to see another PR if there's any further progress on the next steps here.

Issue #142, opened 2019-09-22
How to check if .execute(.) went through ?


gillesdegottex:
Hi,

I've check the doc and the code, though I can't find a way to check if .execute(.) went through.
E.g. with a delete query type, where there is nothing to fetch.

Any idea?

Cheers,

steinitzu:
You mean to get e.g. the row count? I'm not sure if that's exposed, if it is it's probably just the return value of `execute()` (I know it returns the last row id for inserts).

If you mean you just want to know whether the query errored or not, you get an exception if it fails.

dmontagu:
I have a function like this:
```python
async def delete_by_columns_values(
    database: Database, table: sa.Table, columns: AnyCols, values: AnyTuples, pk_columns: List["sa.Column[Any]"]
) -> List[Any]:
    assert all(len(value) == len(columns) for value in values)
    query = table.delete().where(sa.tuple_(*columns).in_(values)).returning(*pk_columns)
    return await database.execute(query)
```
Note: I think the above is postgres-specific -- if I recall correctly, sqlite didn't support this API.

I'm not sure if there is a way to just determine the row count of deleted rows if using sqlite -- it looks like there are special functions to do this in the sqlite package but I'm not sure the functionality is exposed through databases (and at any rate it would probably not be easy to make it portable to other backends).

gillesdegottex:
Yes, sort of a row count.
Though I guess I can rely on the exception.
`execute ()` returns me `0` for delete operations.

tomchristie:
Seconding @dmontagu - you probably want a `RETURNING` clause.
See also #61, #101.

gvbgduh:
As a possible workaround or if you want the full lower level control, there's the allocated interface to access the raw driver as

```
async with database.connection() as connection:
    raw_connection = connection.raw_connection
         status = await raw_connection.copy_records_to_table(...)
```
Then depending on the driver you can utilize its API as is.
But, raw access also means raw response from the DB.

gnat:
Patch featuring the correct behavior is here: https://github.com/encode/databases/pull/150

@gillesdegottex

gillesdegottex:
Thanks a lot !

Issue #61, opened 2019-03-06
Support returning the rowcount for a given query?


orf:
Hey! Thanks for this awesome library.

I'm replacing direct `aiopg` usage with it, but I ran into one issue that I can't seem to solve. When executing a query `aiopg` returns an object with a `.rowcount` attribute. This is quite handy when using an `INSERT ... ON CONFLICT` query to know how many rows have been inserted.

I cannot seem to find out to how expose this information with this library. I'm guessing it might be more of an issue with `asyncpg`, but I was wondering if this is possible and if could be documented somewhere?

tomchristie:
I'm not sure. It'd be worth having a look at `asyncpg` and seeing if it returns something equivalent anywhere. We have just added the ability to drop down to the underlying driver's connection instance. (See #58 and #60) so if `asyncpg` supports it, then you'll be able to do that.

I would also totally be up for having `aiopg` added as an alternate driver too, selected with `postgresql+aiopg://...`, in which case you'd be able to stick with your existing driver if you wanted, and drop down to the raw connection as per #58.

Alternatively, if it looks like we're missing a bit of API that really ought to be provided directly, than I'd be happy to consider whatever we'd need to do to support a row_count in our API.

orf:
I did a bit of digging. This issue is quite enlightening: https://github.com/MagicStack/asyncpg/issues/359

The tl;dr is that `execute()` in `asyncpg` returns the 'status' of the last command executed. This includes the rowcount: https://magicstack.github.io/asyncpg/current/api/index.html#asyncpg.connection.Connection.execute

I'm not sure the best place to fix this. [aiopg doesn't seem to handle it itself](https://github.com/aio-libs/aiopg/search?l=Python&q=rowcount). We could parse the status of the last SQL command that's executed, but to complicate things we don't really expose a way to call `.execute()` without  a raw connection (we only call wrappers like fetch).

Urgh.

tomchristie:
> but to complicate things we don't really expose a way to call .execute() without a raw connection (we only call wrappers like fetch).

Not quite sure what you meant here.

So, our API for `execute()` is currently defined as returning the last cursor id. See https://github.com/encode/databases/blob/master/tests/test_databases.py#L236 for an example.

Are the status messages specific to postgres, or are they supported by mysql/sqlite too?

[...]

PR #150, opened 2019-10-21
Fix: execute() should return rows affected by operation if lastrowid is not set.

gnat:
Patch closes the following issues:
- https://github.com/encode/databases/issues/142
- https://github.com/encode/databases/issues/61

**Old behavior:** Calls to execute() return lastrowid. For some operations this has no benefit.

**New behavior:** Calls to execute() where lastrowid has no benefit will now return rows affected by the operation. (Example: DELETE. So we can now see if a DELETE was successful, and how many rows were deleted). Calls which set a lastrowid value, will still retain the old behavior.

The library encode/databases depends on, aiomysql, returns rows affected, the desired behaviour this patch implements. See: https://github.com/aio-libs/aiomysql/blob/master/aiomysql/cursors.py#L244

### DIFF
--- a/databases/backends/mysql.py
+++ b/databases/backends/mysql.py
@@ -131,7 +131,9 @@ class MySQLConnection(ConnectionBackend):
         cursor = await self._connection.cursor()
         try:
             await cursor.execute(query, args)
-            return cursor.rowcount
+            if cursor.lastrowid == 0:
+                return cursor.rowcount
+            return cursor.lastrowid
         finally:
             await cursor.close()
 
--- a/databases/backends/mysql.py
+++ b/databases/backends/mysql.py
@@ -131,7 +131,7 @@ class MySQLConnection(ConnectionBackend):
         cursor = await self._connection.cursor()
         try:
             await cursor.execute(query, args)
-            return cursor.lastrowid
+            return cursor.rowcount
         finally:
             await cursor.close()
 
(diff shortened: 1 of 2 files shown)


gnat:
I'd like to note that the library you base this on, aiomysql, returns rows affected (rowcount) as well for execute()

See: https://github.com/aio-libs/aiomysql/blob/master/aiomysql/cursors.py#L244

Is it possible the original design decision to return lastrowid may have been an error? @tomchristie

gnat:
@tomchristie Possible to get this merged?

gvbgduh:
Oh, nice! That makes sense. That would also apply for the upcoming `aiopg` backend. 
For `asyncpg` it'll be a bit diff tho, as it returns exactly what postgres returns back, which is of the following form:

_for insert_
```
INSERT oid count
```

_for update_:
```
UPDATE count
```

_for delete_:
```
DELETE count
```

As a plain string, so simple parsing should. Would be very nice to add it to this PR as well, but I can push it a bit later with a different issue to not forget to update `aiopg` when it's merged.

gnat:
@gvbgduh That sounds like a great plan. In the mean time, is there anything holding up getting this merged?

gvbgduh:
@gnat sorry for a delay, merging it now

gnat:
Amazing, thank you @gvbgduh !

PR #156, opened 2019-10-31
Tweak logging

tomchristie:
* Add `INFO` level connect and disconnect messages.
* Change `DEBUG` messages to single line format.
* Support optional colourized formatting. (Not required, won't impact non-supporting log handlers)

Here's an example of how it looks within a uvicorn context...

Screenshot 2019-10-31 at 15 14 47

Note the database connect and disconnect messages at INFO level, and the DEBUG query messages.

### DIFF
--- a/databases/backends/mysql.py
+++ b/databases/backends/mysql.py
@@ -10,7 +10,7 @@ from sqlalchemy.engine.result import ResultMetaData, RowProxy
 from sqlalchemy.sql import ClauseElement
 from sqlalchemy.types import TypeEngine
 
-from databases.core import DatabaseURL
+from databases.core import DatabaseURL, LOG_EXTRA
 from databases.interfaces import ConnectionBackend, DatabaseBackend, TransactionBackend
 
 logger = logging.getLogger("databases")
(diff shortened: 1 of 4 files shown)


tomchristie:
Connect/disconnect now highlights the database URL (password obscured).

Allows us to present really nice clear startup info...

Screenshot 2019-10-31 at 15 26 46

pvanliefland:
Great! Just one thought: if I understand correctly, the "Connected to database" means something like "Established connection pool" (except for SQLite, where nothing happens at connection time).

Don't you think it would be useful as well to have logs when a connection is acquired from the pool?

tomchristie:
Potentially yeah. We might consider using TRACE logging level for anything at that kind of level. Certainly transactions and savepoints should be logged at the same DEBUG level as other queries tho. (Maybe they already are - haven’t double checked?)

gvbgduh:
@pvanliefland if I'm not wrong (and at least in asyncpg) the actual connection is done _by the pool_ and seemed that in databases those connections are populated ahead (with the min number) and the pool is rather manager. (But, asyncpg has a particular implementation of that and I'm not 100% sure about other backends.)
So, I'm under impression that that logline appears when the pool is populated with created connections, but more detailed info is probably indeed for the trace level.

PR #158, opened 2019-11-03
Delay global_connection and global_transaction initialization to latest possible moment

### LINKED ISSUE #157: _query_lock implementation causes "Future attached to a different loop" error when gather() is used with global Database(force_rollback=True) instance in app tests
This is tangentially related to #125

When writing tests for my project I've found that following [Starlette's database usage docs](https://www.starlette.io/database/) and [Databases tests docs](https://www.encode.io/databases/tests_and_migrations/) in app results in setup that makes it impossible to test application logic that uses `gather` to initialize multiple queries at single time.

## Reproduction

Starlette's database example directs developers to create single `Database` instance for their application:

```python
database = databases.Database(DATABASE_URL)

app = Starlette()

app.add_event_handler("startup", database.connect)
app.add_event_handler("shutdown", database.disconnect)
```

However in tests you would like to enforce some sort of isolation between tests. Easy way to achieve this is to run all tests in isolation. Hence `force_rollback` flag on `Database`. If developer updates their implementation, they will end with something like this:

```python
if TESTING:
    database = Database(TEST_DATABASE_URL, force_rollback=True)
else:
    database = Database(DATABASE_URL)

app = Starlette()

app.add_event_handler("startup", database.connect)
app.add_event_handler("shutdown", database.disconnect)
```

This code will run fine until we try to use `gather` somewhere in the logic and write a test for it:

```python
from asyncio import gather

from .database import database


async def logic_using_gather():
    return await gather(
        database.fetch_one("SELECT pg_sleep(1)"),
        database.fetch_one("SELECT pg_sleep(1)"),
    )


@pytest.fixture(scope="function")
async def db():
    async with database:
        yield


@pytest.mark.asyncio
async def test_my_logic(db):
    await logic_using_gather()
``` 

We will get an error:

```
RuntimeError: Task <Task pending coro=<Database.execute() running at /usr/local/lib/python3.7/site-packages/databases/core.py:122> cb=[gather.<locals>._done_callback() at /usr/local/lib/python3.7/asyncio/tasks.py:691]> got Future <Future pending> attached to a different loop
```

## The isssue

When `Database` is initialized with `force_rollback=True`, it internally creates single "global connection" object which in turn internally creates `asyncio.Lock` instance for itself.

`asyncio.Lock` can be initialized with `loop` argument, but if none is provided, it will try to get current running event loop, or create new event loop is none exists. The former is the case when app is running behind one of ASGI servers that provide event loop, while latter is the case in the test setups.

Error is result of test-runner provided event loop running task which itself tries to `await asyncio.Lock()` that was initialized with its own event loop. This is not the case when queries are ran one after another, because first query will release the lock before next one is sent to event loop for execution:

```
await database.fetch_one("SELECT pg_sleep(1)")
await database.fetch_one("SELECT pg_sleep(1)")
```

However when `gather` is used, both tasks are sent to event loop at same time, and because `_query_lock._locked` is already set to True, `await _query_lock` will cause `asyncio.Lock` to create and send waiting task to event loop. This task will be sent to different event loop than one running rest of the logic, producing the resulting error.

## Reduced example

Reaching back to the discussion in #125, its pointed out that there's a test in databases for using `gather` to await multiple queries, but this test checks unrealistic use of `Database` together with `gather`:

```python
@async_adapter
async def test_concurrent_access_on_single_connection(database_url):
    async with Database(database_url, force_rollback=True) as database:

        async def db_lookup():
            if str(database_url).startswith("postgresql"):
                await database.fetch_one("SELECT pg_sleep(1)")

        await asyncio.gather(db_lookup(), db_lookup())
```

Whereas more real (and failing) usage would look like this:

```python
def test_concurrent_access_on_single_connection(database_url):
    database = Database(database_url, force_rollback=True)

    @async_adapter
    async def run_db_lookups():
        async with database:
            async def db_lookup():
                if str(database_url).startswith("postgresql"):
                    await database.fetch_one("SELECT pg_sleep(1)")

            await asyncio.gather(db_lookup(), db_lookup())
    
    run_db_lookups()
```

Above test will fail with same error because now `database` and `async_adapter` will use two different event loops for their logic.

rafalp:
This PR changes `Database` implementation so global connection is created only when its actually needed: when connection with backend is initialized. That way connection and its locks are initialized within same event loop, preventing issue documented in #157.

I've also had to change async_adapter implementation so it creates new event loop on every use instead of reusing already-existing event loop, which allowed me to reproduce error described in #157. It should also make tests run in env closer to "real" usage.

### DIFF
--- a/databases/core.py
+++ b/databases/core.py
@@ -72,12 +72,6 @@ class Database:
         self._global_connection = None  # type: typing.Optional[Connection]
         self._global_transaction = None  # type: typing.Optional[Transaction]
 
-        if self._force_rollback:
-            self._global_connection = Connection(self._backend)
-            self._global_transaction = self._global_connection.transaction(
-                force_rollback=True
-            )
-
     async def connect(self) -> None:
         """
         Establish the connection pool.
@@ -91,7 +85,14 @@ class Database:
         self.is_connected = True
 
         if self._force_rollback:
-            assert self._global_transaction is not None
+            assert self._global_connection is None
+            assert self._global_transaction is None
+
+            self._global_connection = Connection(self._backend)
+            self._global_transaction = self._global_connection.transaction(
+                force_rollback=True
+            )
+
             await self._global_transaction.__aenter__()
 
     async def disconnect(self) -> None:
@@ -101,8 +102,13 @@ class Database:
         assert self.is_connected, "Already disconnected."
 
         if self._force_rollback:
+            assert self._global_connection is not None
             assert self._global_transaction is not None
+
             await self._global_transaction.__aexit__()
+
+            self._global_transaction = None
+            self._global_connection = None
 
         await self._backend.disconnect()
         logger.info(
--- a/tests/test_databases.py
+++ b/tests/test_databases.py
@@ -100,7 +100,7 @@ def async_adapter(wrapped_func):
 
     @functools.wraps(wrapped_func)
     def run_sync(*args, **kwargs):
-        loop = asyncio.get_event_loop()
+        loop = asyncio.new_event_loop()
         task = wrapped_func(*args, **kwargs)
         return loop.run_until_complete(task)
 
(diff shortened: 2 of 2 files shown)


tomchristie:
Seems reasonable yup - what's the behavior of the test case without this change?

rafalp:
@tomchristie if you were to merge only tests part of this PR, then test case would reproduce my initial issue, and tests would fail with `RuntimeError: Task <Task pending coro=<Database.execute() running at /usr/local/lib/python3.7/site-packages/databases/core.py:122> cb=[gather.<locals>._done_callback() at /usr/local/lib/python3.7/asyncio/tasks.py:691]> got Future <Future pending> attached to a different loop` error.

tomchristie:
Looks all v sensible to me.
I'll leave this for anyone else on the maintainance team to merge, then we should start looking at adding more PyPI permissions for releases.

gvbgduh:
Looks safe to merge anyway!

But for future we can revise the `async_adapter` itself and async clients, I'm under impresion it's a bit more convenient to run everything within one event loop (or like one loop per worker/process), which is generally provided by `pytest.mark.asyncio` in tests.
There was some confusion with the `starlette`'s `test_client` that creates the conflicting event loop. Also some confusion in scenarious, running `databases` independently or within the service driven by `starlette`.

PR #165, opened 2019-12-15
Add scheme for the postgresql backend

graphman65:
Add "postgres" as a valid scheme for the postgresql backend.

See https://www.postgresql.org/docs/10/libpq-connect.html#id-1.7.3.8.3.6


NJannasch:
Issue raised in: https://github.com/encode/databases/issues/146
Fix already available in: https://github.com/encode/databases/pull/147

gvbgduh:
Merged in encode/databases#147

PR #173, opened 2020-03-13
Make sure the correct values are returned from a PostgreSQL record

rzane:
It's fairly common to select two columns that are assigned the same name. In those scenarios, you can't really rely on the dictionary version of the record, but you get the values by their index.

Previously, when two columns with the same name were selected, the PostgreSQL Record would return only the first column's value.

```python
>>> query = "SELECT 1 AS id, 2 AS id"
>>> row = await database.fetch_one(query=query)
>>> list(row.values())
[1, 1]
```

This PR fixes that bug and returns the correct values.


rzane:
The build appears to be failing because coverage isn't 100%, but I don't quite see how this PR could have caused a decline in coverage?

tomchristie:
Fantastic, thanks!

@rzane Since this is your first approved PR, you're welcome to an invite to `maintainers` if you'd like it? That'd give you permissions to approve or merge PRs.

Thanks for your time on this. 😀

vmarkovtsev:
@tomchristie did you mean myself or Ray?..

tomchristie:
I meant Ray, yup. We're trying to move to an open doors policy of "invites to the maintainers team for anyone who gets a PR through"

vmarkovtsev:
All right :+1:  I wonder if https://github.com/encode/databases/pull/199 counts in my case, too.

tomchristie:
Yup absolutely, just not always succeeding in tracking it! 😬

PR #193, opened 2020-04-24
Speed up PostgresConnection fetch() and iterate()

vmarkovtsev:
I noticed that the column maps in the Postgres abstraction can moved outside of the loop. According to the profiling results, this makes -25% less overhead on wrapping `asyncpg.Record`.

### DIFF
--- a/databases/backends/postgres.py
+++ b/databases/backends/postgres.py
@@ -74,25 +74,30 @@ class PostgresBackend(DatabaseBackend):
 
 
 class Record(Mapping):
+    __slots__ = (
+        "_row",
+        "_result_columns",
+        "_dialect",
+        "_column_map",
+        "_column_map_int",
+        "_column_map_full",
+    )
+
     def __init__(
-        self, row: asyncpg.Record, result_columns: tuple, dialect: Dialect
+        self,
+        row: asyncpg.Record,
+        result_columns: tuple,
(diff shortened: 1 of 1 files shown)


databases/backends/postgres.py
```diff
@@ -201,6 +214,21 @@ def _compile(self, query: ClauseElement) -> typing.Tuple[str, list, tuple]:
...
         return compiled_query, args, compiled._result_columns
 
+    @staticmethod
+    def _create_column_maps(
+        result_columns: tuple,
+    ) -> typing.Tuple[
+        typing.Mapping[typing.Any, typing.Tuple[int, TypeEngine]],
+        typing.Mapping[int, typing.Tuple[int, TypeEngine]],
+        typing.Mapping[str, typing.Tuple[int, TypeEngine]],
+    ]:
```
tomchristie, line 224: Okay. I'd be quite keen for us to document the return types here, maybe?

I realise that's not *currently* the case, but if we're touching this part of the codebase, then now might be a good time for us to improve that, since it's a really unclear area at the moment.

vmarkovtsev: Done

vmarkovtsev:
@tomchristie How about now? I'd rather add the rest of the docs later.

PR #199, opened 2020-04-28
Build queries before acquiring the lock

vmarkovtsev:
Building the query takes some time, especially if it is big. We hit real performance problems in prod because of that (hundreds of bound parameters, etc.).


vmarkovtsev:
The checks are actually successful, that's the GitHub's CI internals which are not.

tomchristie:
@vmarkovtsev Yuck. Okay, let's see if we can re-run that.

Having just moved off travis, I'm *really* hoping GitHub actions is decently robust.

Issue #198, opened 2020-04-27
Why require psycopg for databases[postgresql]?


retnikt:
When installing databases with asyncpg PostgreSQL support using `pip install databases[postgresql]`, `psycopg2-binary` is installed as well. This is [not recommended](https://www.psycopg.org/docs/install.html#binary-install-from-pypi), and seems also unnecessary given that there are no references to it in the code, and asyncpg does not require it. Is there any reason for including it as a dependency or is it just a relic of an older version?

fullonic:
The need of psycopg2 is related with sqlalchemy, which databases depends on. But you are right, it shouldn't be the binary version.

tomchristie:
We could consider removing it entirely. (And the corresponding mysql sync driver too.)

`databases` itself *doesn't* require them, but you *do* need them if you want to do `engine.create_all`, or if you want to use `alembic` for database migrations. (In practice you're almost certainly going to want to do one of those.)

I guess a first step here might be - we'll accept a pull request switching it to the non-binary version, and we can follow up if there's any problems with that.

jpic:
Not sure if this helps, but async create_all is in SQLA 1.4b1:

```
with engine.begin() as conn:         
    await conn.run_sync(metadata.create_all)
```

Maybe that'll also be an option.

PR #204, opened 2020-05-01
Replace psycopg2-binary with psycopg2 (#198)

retnikt:


JayH5:
Just a drive-by comment that this change probably needs a bit of a warning in the changelog as the `psycopg2` package requires build tools, `libpq`, etc, whereas before the `psycopg2-binary` package did not.

tomchristie:
Okay, which makes me want to re-evaluate what benefit this change would give us.

retnikt:
Well, really I think users should have the choice between the binary or non-binary version depending on their circumstances. The only option for this I believe would be to remove the requirement entirely and tell users in the docs that one of them is required for table creation/migrations. This also allows the flexibility of running your migrations entirely separately from your web server.

tomchristie:
Then let's drop `psycopg2` and `pymysql` entirely, and make sure that the next version is a proper version bump, and calls out the requirement to install a sync driver if you're using `alembic` or SQL Alchemy's `engine.create_all()`

vmarkovtsev:
Here I am, the one who released 0.4.0 and broke my own production builds because of this :sweat:

So shall I remove these requirements completely then?

PR #206, opened 2020-05-04
Implement column defaults for INSERT/UPDATE

GarbageHamburger:
This PR implements usage of column defaults from SQLAlchemy. In regular SQLAlchemy the default column values are fetched using `ExecutionContext._exec_default`. However, we don't use the `ExecutionContext` here, so a function is implemented to take its place for column defaults.

### DIFF
--- a/databases/core.py
+++ b/databases/core.py
@@ -9,6 +9,9 @@ from urllib.parse import SplitResult, parse_qsl, urlsplit
 
 from sqlalchemy import text
 from sqlalchemy.sql import ClauseElement
+from sqlalchemy.sql.dml import ValuesBase
+from sqlalchemy.sql.expression import type_coerce
+
 
 from databases.importer import import_from_string
 from databases.interfaces import ConnectionBackend, DatabaseBackend, TransactionBackend
@@ -294,10 +297,50 @@ class Connection:
             query = text(query)
 
             return query.bindparams(**values) if values is not None else query
-        elif values:
+
+        # 2 paths where we apply column defaults:
+        # - values are supplied (the object must be a ValuesBase)
+        # - values is None but the object is a ValuesBase
+        if values is not None and not isinstance(query, ValuesBase):
+            raise TypeError("values supplied but query doesn't support .values()")
+
+        if values is not None or isinstance(query, ValuesBase):
+            values = Connection._apply_column_defaults(query, values)
             return query.values(**values)
 
         return query
+
+    @staticmethod
+    def _apply_column_defaults(query: ValuesBase, values: dict = None) -> dict:
+        """Add default values from the table of a query."""
+        new_values = {}
+        values = values or {}
+
+        for column in query.table.c:
+            if column.name in values:
+                continue
+
+            if column.default:
+                default = column.default
+
+                if default.is_sequence:  # pragma: no cover
+                    # TODO: support sequences
+                    continue
+                elif default.is_callable:
+                    value = default.arg(FakeExecutionContext())
+                elif default.is_clause_element:  # pragma: no cover
+                    # TODO: implement clause element
+                    # For this, the _build_query method needs to
+                    # become an instance method so that it can access
+                    # self._connection.
+                    continue
+                else:
+                    value = default.arg
+
+                new_values[column.name] = value
+
+        new_values.update(values)
+        return new_values
 
 
 class Transaction:
@@ -489,3 +532,20 @@ class DatabaseURL:
 
     def __eq__(self, other: typing.Any) -> bool:
         return str(self) == str(other)
+
+
+class FakeExecutionContext:
+    """
+    This is an object that raises an error when one of its properties are
+    attempted to be accessed. Because we're not _really_ using SQLAlchemy
+    (besides using its query builder), we can't pass a real ExecutionContext
+    to ColumnDefault objects. This class makes it so that any attempts to
+    access the execution context argument by a column default callable
+    blows up loudly and clearly.
+    """
+
+    def __getattr__(self, _: str) -> typing.NoReturn:  # pragma: no cover
+        raise NotImplementedError(
+            "Databases does not have a real SQLAlchemy ExecutionContext "
+            "implementation."
+        )
--- a/tests/test_databases.py
+++ b/tests/test_databases.py
@@ -70,6 +70,17 @@ prices = sqlalchemy.Table(
     sqlalchemy.Column("price", sqlalchemy.Numeric(precision=30, scale=20)),
 )
 
+# Used to test column default values
+timestamps = sqlalchemy.Table(
+    "timestamps",
+    metadata,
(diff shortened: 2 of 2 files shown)


databases/core.py
```diff
@@ -9,6 +9,9 @@
 
 from sqlalchemy import text
 from sqlalchemy.sql import ClauseElement
+from sqlalchemy.sql.dml import ValuesBase
+from sqlalchemy.sql.expression import type_coerce
+
```
rafalp, line 14:
```suggestion
```

Double blank line

GarbageHamburger:
Any updates on this?

tomchristie:
It's difficult because I'm not really sure if we want this functionality or not.

andre-dasilva:
@tomchristie I am just curious. Why would you not want this functionality? I am currently writing a application with fastapi and encode/databases and i want the timestamp to be created on insert. currently i don't have a way of doing this.

camgunz:
It also seems necessary for UUID primary keys, which are pretty common these days.

vmarkovtsev:
The column defaults are higher level sqla API ("Query") than the API used in `databases` ("Core"). Emulating them properly is quite hard. For example, the functions can access the context in several ways, and filling it is twisted.

The problem can be solved outside of  `databases` package by adding a new method to your model base class to explicitly create defaults. 

There is another way to access the context:

GarbageHamburger:
@vmarkovtsev SQLAlchemy core does have defaults, see [the `default` parameter for `Column.__init__`](https://docs.sqlalchemy.org/en/13/core/metadata.html#sqlalchemy.schema.Column.params.default). I agree that correctly emulating defaults is hard due to the types of things a column default can be (it can be a `ClauseElement`!), but that's no reason for it to not be included.

ricosuave73:
+1 for this. I followed the FastAPI tutorial and tried to implement a UUID for sqlite. I couldn't figure out why it didn't work until I specifically checked this repo for issues around the defaults.
And defaults are indeed part of sqlalchemy core.

ElPapi42:
This is a must, thanks

ElPapi42:
Approve this pls

vmarkovtsev (changes requested):
Sorry fellows, I wish I approved this, but this looks too hacky right now to be included as a general solution.

databases/core.py
```diff
@@ -294,11 +296,51 @@ def _build_query(
...
-        elif values:
+
+        # 2 paths where we apply column defaults:
+        # - values are supplied (the object must be a ValuesBase)
+        # - values is None but the object is a ValuesBase
+        if values is not None and not isinstance(query, ValuesBase):
+            raise TypeError("values supplied but query doesn't support .values()")
+
+        if values is not None or isinstance(query, ValuesBase):
+            values = Connection._apply_column_defaults(query, values)
```
vmarkovtsev, line 307:
```suggestion
        elif values:
            assert isinstance(query, ValuesBase)
            values = self._apply_column_defaults(query, values)
```

databases/core.py
```diff
@@ -294,11 +296,51 @@ def _build_query(
...
+        for column in query.table.c:
+            if column.name in values:
+                continue
+
+            if column.default:
+                default = column.default
+
+                if default.is_sequence:  # pragma: no cover
+                    # TODO: support sequences
+                    continue
```
vmarkovtsev, line 327: assert not default.is_sequence, "sequences are not supported, PRs welcome"

databases/core.py
```diff
@@ -294,11 +296,51 @@ def _build_query(
...
         return query
 
+    @staticmethod
+    def _apply_column_defaults(query: ValuesBase, values: dict = None) -> dict:
+        """Add default values from the table of a query."""
+        new_values = {}
+        values = values or {}
+
+        for column in query.table.c:
+            if column.name in values:
```
vmarkovtsev, line 319: `values` can have columns as keys, in such cases this will be always false

databases/core.py
```diff
@@ -294,11 +296,51 @@ def _build_query(
...
+
+            if column.default:
+                default = column.default
+
+                if default.is_sequence:  # pragma: no cover
+                    # TODO: support sequences
+                    continue
+                elif default.is_callable:
+                    value = default.arg(FakeExecutionContext())
+                elif default.is_clause_element:  # pragma: no cover
```
vmarkovtsev, line 330:
```suggestion
                assert not default.is_clause_element, "clause defaults are not supported, PRs welcome"
```

vmarkovtsev: + can you please group assertions together

databases/core.py
```diff
@@ -294,11 +296,51 @@ def _build_query(
...
+                    # TODO: implement clause element
+                    # For this, the _build_query method needs to
+                    # become an instance method so that it can access
+                    # self._connection.
+                    continue
+                else:
+                    value = default.arg
+
```
vmarkovtsev, line 338:
```suggestion
```

databases/core.py
```diff
@@ -489,3 +531,20 @@ def __repr__(self) -> str:
...
+    """
+    This is an object that raises an error when one of its properties are
+    attempted to be accessed. Because we're not _really_ using SQLAlchemy
+    (besides using its query builder), we can't pass a real ExecutionContext
+    to ColumnDefault objects. This class makes it so that any attempts to
+    access the execution context argument by a column default callable
+    blows up loudly and clearly.
+    """
+
+    def __getattr__(self, _: str) -> typing.NoReturn:  # pragma: no cover
```
vmarkovtsev, line 546: We should cover this by another test which tests raising NotImplementedError

databases/core.py
```diff
@@ -489,3 +531,20 @@ def __repr__(self) -> str:
 
     def __eq__(self, other: typing.Any) -> bool:
         return str(self) == str(other)
+
+
+class FakeExecutionContext:
```
vmarkovtsev, line 536:
```suggestion
class DummyExecutionContext:
```
(it's not completely fake, after all)

databases/core.py
```diff
@@ -489,3 +531,20 @@ def __repr__(self) -> str:
...
+    This is an object that raises an error when one of its properties are
+    attempted to be accessed. Because we're not _really_ using SQLAlchemy
+    (besides using its query builder), we can't pass a real ExecutionContext
+    to ColumnDefault objects. This class makes it so that any attempts to
+    access the execution context argument by a column default callable
+    blows up loudly and clearly.
+    """
+
+    def __getattr__(self, _: str) -> typing.NoReturn:  # pragma: no cover
+        raise NotImplementedError(
```
vmarkovtsev, line 547: btw: my own custom implementation of this (yep, I have a similar hack in my prod), I pass the current `values` to this context so that it becomes much more useful.

vmarkovtsev (changes requested):
These are my suggestions :+1:

etscript:
Any updates on this?

PR #211, opened 2020-05-28
Avoid dangling task-local connections after Database.disconnect()

vmarkovtsev:
This change forces initialization of a new Connection if we connect() again.

### DIFF
--- a/databases/core.py
+++ b/databases/core.py
@@ -119,7 +119,7 @@ class Database:
             self._global_transaction = None
             self._global_connection = None
         else:
-            self._connection_context = ContextVar("connection_context")
+            self._connection_context = contextvars.ContextVar("connection_context")
 
         await self._backend.disconnect()
         logger.info(
--- a/databases/core.py
+++ b/databases/core.py
@@ -112,6 +112,8 @@ class Database:
 
             self._global_transaction = None
             self._global_connection = None
+        else:
+            self._connection_context = ContextVar("connection_context")
 
         await self._backend.disconnect()
         logger.info(


tomchristie:
Heya! Thanks for the PR here. I think I'd need a need clearer background here...
Even if it's too awkward to write a direct test case for this, it'd be good to walk through how we can demonstrate the issue?

vmarkovtsev:
Sure, here is what happens. Let's suppose we've got `db = databases.Database(...)`.

1. `await db.connect()`.
2. Run some query in some coroutine. `db._connection_context` sets to the task-local `Connection` in that coroutine context.
3. `await db.disconnect()`
4. We cannot `await db.connect()` anymore because `db._connection_context` is still pointing at the old `Connection` with a broken `_backend`, so the subsequent `db` queries are wrong.
5. We continue referencing the connections for all the coroutine contexts that we touched.

The current code cleans up the global connection which we manipulate in case `force_rollback` is `True`. There is no simialr cleanup of the task-local connections.

vmarkovtsev:
@tomchristie Ping

vmarkovtsev:
@tomchristie ping

vmarkovtsev:
@tomchristie ping
(sorry for so many pings, I know how easy it is to miss a GitHub notification/PR review request, feel free to stop the pings :+1:)

vmarkovtsev:
@tomchristie friendly ping

taybin:
I've been seeing "Connection is not acquired" assertion errors in my long running processes.  Could this be a cause of that?

aminalaee:
@taybin I'm not sure if this can be related to your issue.If you have more info about your issue, please open a new one.
Thanks for bringing this up anyway :)

aminalaee (approved):

taybin:
@aminalaee On further research, I believe #283 is the cause of what I've been seeing.

aminalaee:
@taybin That one looks good too. I'll check it again.

PR #212, opened 2020-06-02
Update connections_and_transactions.md

Kludex:
Fix typo: `else` -> `finally`

### DIFF
--- a/docs/connections_and_transactions.md
+++ b/docs/connections_and_transactions.md
@@ -74,7 +74,7 @@ try:
     ...
 except:
     transaction.rollback()
-else:
+finally:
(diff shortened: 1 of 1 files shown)


docs/connections_and_transactions.md
```diff
@@ -74,7 +74,7 @@ try:
     ...
 except:
     transaction.rollback()
-else:
+finally:
```
vmarkovtsev, line 77: There is no problem here. See: https://github.com/encode/databases/blob/master/databases/core.py#L328

If you fail the transaction, you never commit it.

Kludex:
Thank you :)

PR #213, opened 2020-06-02
Add `__str__` method to postgresql Record class

skewty:
I know this PR wasn't discussed yet but it is so minimal I didn't think it would be worth it.

### DIFF
--- a/databases/backends/postgres.py
+++ b/databases/backends/postgres.py
@@ -132,7 +132,7 @@ class Record(Mapping):
         return self._row.__str__()
 
     def __repr__(self) -> str:
-        return self._row.__repr__()
+        return f"Record({repr(self._row)}, {repr(self._result_columns)}, {repr(self._dialect)})"
 
 
 class PostgresConnection(ConnectionBackend):
--- a/databases/backends/postgres.py
+++ b/databases/backends/postgres.py
@@ -128,6 +128,9 @@ class Record(Mapping):
     def __len__(self) -> int:
         return len(self._row)
 
+    def __str__(self) -> str:
+        return f"""{{{", ".join(f"'{key}': {value!r}" for key, value in self.items())}}}"""
+
 
 class PostgresConnection(ConnectionBackend):
     def __init__(self, database: PostgresBackend, dialect: Dialect):
--- a/databases/backends/postgres.py
+++ b/databases/backends/postgres.py
@@ -129,7 +129,10 @@ class Record(Mapping):
         return len(self._row)
 
     def __str__(self) -> str:
-        return f"""{{{", ".join(f"'{key}': {value!r}" for key, value in self.items())}}}"""
+        return self._row.__str__()
+
+    def __repr__(self) -> str:
+        return self._row.__repr__()
 
 
 class PostgresConnection(ConnectionBackend):


skewty:
https://github.com/encode/databases/issues/214

databases/backends/postgres.py
```diff
@@ -128,6 +128,9 @@ def __iter__(self) -> typing.Iterator:
     def __len__(self) -> int:
         return len(self._row)
 
+    def __str__(self) -> str:
+        return f"""{{{", ".join(f"'{key}': {value!r}" for key, value in self.items())}}}"""
```
vmarkovtsev, line 132: How about
```suggestion
        return str(dict(self))
```

The result should be the same.

skewty: @vmarkovtsev imagine this scenario: 

A long list of Records is generated.  Each dict will be allocated and will quickly go out of scope but will remain until garbage collected. You create a spike in memory allocation which could cripple a low memory device (think embedded) and force it to swap.  

A long list of Records is a common and many IDEs will automatically call `__str__` on entries to show something useful in the debugger so I would imagine this scenario is not uncommon.

skewty: > I am not familiar with the other DB backends - do they have similar __str__-s?

I wasn't concerned about other libraries when I wrote that method.  To me a typing.Mapping is better presented as a string similar to CPython's `dict.__str__()`.  It certainly makes more sense that mimicking CPython's `tuple.__str__()`.

Imagine this scenario.  Your row contains 12 integers.  If you can't remember the query string used, and the `tuple.__str__()` was mimicked you would likely have difficulty identifying which in is the one you are interested in.  Conversely, with the field names also supplied, the one you are interested in is obvious.

vmarkovtsev, line 132: Python counts references, so the dict will not be garbage collected. It will disappear immediately after exiting the visibility scope. Garbage collection does a different thing.

OK, then I have this suggestion:

```suggestion
    def __repr__(self) -> str:
        return f"Record({repr(self._row)}, {repr(self._result_columns)}, {repr(self._dialect)})"
```

This way the IDE is happy (it will use `__repr__` if `__str__` does not exist), and we leverage the Cython implementation of `asyncpg.Record.__repr__` which is potentially faster.

skewty:
## Test Memory Allocation

```py
import tracemalloc
import linecache
import os
import gc
import time
import uuid

def display_top(snapshot, key_type='lineno', limit=1):
    """Source: https://stackoverflow.com/questions/552744/how-do-i-profile-memory-usage-in-python"""

    snapshot = snapshot.filter_traces((
        tracemalloc.Filter(False, "<frozen importlib._bootstrap>"),
        tracemalloc.Filter(False, "<unknown>"),
    ))
    top_stats = snapshot.statistics(key_type)

    print("Top %s lines" % limit)
    for index, stat in enumerate(top_stats[:limit], 1):
        frame = stat.traceback[0]
        # replace "/path/to/module/file.py" with "module/file.py"
        filename = os.sep.join(frame.filename.split(os.sep)[-2:])
        print("#%s: %s:%s: %.1f KiB"
              % (index, filename, frame.lineno, stat.size / 1024))
        line = linecache.getline(frame.filename, frame.lineno).strip()
        if line:
            print('    %s' % line)

    other = top_stats[limit:]
    if other:
        size = sum(stat.size for stat in other)
        print("%s other: %.1f KiB" % (len(other), size / 1024))
    total = sum(stat.size for stat in top_stats)
    print("Total allocated size: %.1f KiB" % (total / 1024))

class Record(dict):
    def __str_a__(self) -> str:
        return f"""{{{", ".join(f"'{key}': {value!r}" for key, value in self.items())}}}"""

    def __str_b__(self) -> str:
        return str(dict(self))


records = [Record(id=i, timestamp=time.time(), uuid=str(uuid.uuid4())) for i in range(100000)]

gc.collect()
gc.disable()

tracemalloc.start()
for record in records:
    record.__str_a__()
snapshot_a = tracemalloc.take_snapshot()

gc.enable()
gc.collect()
gc.disable()

tracemalloc.start()
for record in records:
    record.__str_b__()
snapshot_b = tracemalloc.take_snapshot()

print("\n\n")
print("="*60)
print("Snapshot A")
print("="*60)
display_top(snapshot_a)

linecache.clearcache()

print("\n\n")
print("="*60)
print("Snapshot B")
print("="*60)
display_top(snapshot_b)
print("="*60)
```

### Results

```shell
============================================================
Snapshot A
============================================================
Top 1 lines
#1: scratches/scratch_5.py:37: 0.6 KiB
    return f"""{{{", ".join(f"'{key}': {value!r}" for key, value in self.items())}}}"""
2 other: 0.9 KiB
Total allocated size: 1.5 KiB

============================================================
Snapshot B
============================================================
Top 1 lines
#1: scratches/scratch_5.py:40: 6.6 KiB
    return str(dict(self))
8 other: 3.2 KiB
Total allocated size: 9.7 KiB
```

## Conclusion

`__str_a__()` allocated 0.6 KiB and `__str_b__()` allocated 6.6 KiB so i'd say `__str_a__` is roughly 10x more memory efficient running the code above.  If my understanding is correct the gap will further increase with the amount of data in the Record.

I am not emotionally tied to either solution; I'd just like one upstream :)     SQLite records already have a `__str__` / `__repr__` so I was surprised that the Postgresql records did not.

skewty:
@vmarkovtsev good idea about checking if postgres's upstream Record class has a `__repr__` or `__str__` class.  Like you commented it could be a very efficient C/C++ algorithm.

vmarkovtsev:
I checked before posting the comment: it seems to have `__repr__` but it doesn't have `__str__`.

Regarding Python GC: here is a post that describes the difference between reference counting and GC: https://rushter.com/blog/python-garbage-collector/
Nice memory benchmark! Yep, f-strings allocate less memory, that's undoubted. So let's try delegating to `asyncpg.Record`, if it does not work out - go with f-strings.

skewty:
The `self._row.__repr__()` method produces an output that already includes the column names so adding `self._result_columns` parameter to the repr isn't really adding anything useful.

Example output with two columns:   `"<Record id_label='system' gui_label='System'>"`
so it is close to what the f-string method (above) was generating.

Did you want to include `self._result_columns` because it more closely follows the `__init__()` parameters (more pythonic)?

Also notice that `self._row.__repr__()` does not join the field+value pairs with a comma so it isn't really a python repr() and is more of a str() anyways.

vmarkovtsev:
Sorry for making you wait!

> FAIL Required test coverage of 100% not reached. Total coverage: 99.88%

Can you please add trivial `str` and `repr` tests? I'll merge at once then.

databases/backends/postgres.py
```diff
@@ -128,6 +128,12 @@ def __iter__(self) -> typing.Iterator:
     def __len__(self) -> int:
         return len(self._row)
 
+    def __str__(self) -> str:
+        return self._row.__str__()
+
+    def __repr__(self) -> str:
+        return self._row.__repr__()
```
vmarkovtsev, line 135: I am not insisting here but this goes closer to the nature of `repr`:

```suggestion
        return f"Record({repr(self._row)}, {repr(self._result_columns)}, {repr(self._dialect)})"
```

AlJohri:
gentle ping! just started using `databases` today and noticed the `__str__` wouldn't show me anything useful unless I ran `print(rows[0]._row)`

any blockers on merging this PR?

vmarkovtsev:
@skewty Ping from me, too :smile: 

@AlJohri Please feel free to finish this PR! As you see there is not much left to do.

skewty:
Working on unit testing this is more trouble than I had anticipated with the embedded calls to `repr(self._row)` and `repr(self._dialect)`.

Am I okay to assume the project root is in PYTHONPATH?  I ask because I want to import fixtures, constants and such used by other unit tests as below.

Given `with database.transaction(force_rollback=True)` the `force_rollback=True` will essentially do the db call and then roll it back?  This was put in place purely for testing I presume.  Or am I understanding it wrong?  Also, does it therefore follow that it is safe to assume the database is empty when my test method is called?

Here is the start of what I have in my fork in `tests/test_postgres_record.py`:

```python
import pytest

from databases import Database, DatabaseURL
from databases.backends.postgres import Record
from tests.test_databases import DATABASE_URLS, async_adapter, notes


@async_adapter
async def test_postgres_record_repr(database_url):
    database_url = DatabaseURL(database_url)
    if database_url.dialect != "postgresql":
        pytest.skip("Test requires postgresql dialect")

        async with database.transaction(force_rollback=True):
            query = notes.insert()
            await database.execute(query, values)

            # can I assume the row I just inserted is the only one in the database?
```

PR #237, opened 2020-08-20
Avoid commit errors and fail to disconnect from the database.

Z-2020:
For example, when the database is set up with optimistic lock,
the commit will give an error and the connection cannot be broken.

### DIFF
--- a/databases/core.py
+++ b/databases/core.py
@@ -7,6 +7,7 @@ import typing
 from types import TracebackType
 from urllib.parse import SplitResult, parse_qsl, urlsplit
 
+from pymysql.err import IntegrityError
 from sqlalchemy import text
 from sqlalchemy.sql import ClauseElement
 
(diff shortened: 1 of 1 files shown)


databases/core.py
```diff
@@ -363,7 +364,11 @@ async def commit(self) -> None:
         async with self._connection._transaction_lock:
             assert self._connection._transaction_stack[-1] is self
             self._connection._transaction_stack.pop()
-            await self._transaction.commit()
+            try:
+                await self._transaction.commit()
```
vmarkovtsev, line 368: (just a question) will the following work?
```python
try:
    transaction.commit()
except IntegrityError:
    transaction.rollback()
```
That is, catching it in the user code instead.

Z-2020: > (just a question) will the following work?
> 
> ```python
> try:
>     transaction.commit()
> except IntegrityError:
>     transaction.rollback()
> ```
> 
> That is, catching it in the user code instead.

async def **rollback**(self) -> None:
        async with self._connection._transaction_lock:
            assert **self._connection._transaction_stack[-1]** is self  # Here _transaction_stack will be an empty list.
           ......

when I use **transaction.start()** before, unfortunately, TIDB did not support SAVEPOINT.
May I ask if you have any other suggestions? 

I look forward to your reply!

vmarkovtsev:
@Z-2020 can I close this?

vmarkovtsev:
Feel free to ping me if there is something left.

PR #242, opened 2020-09-12
Allowing extra transaction options

kpdemetriou:
These changes make it possible to use advanced transaction options (currently only with asyncpg) such as different isolation levels or access modes.


kpdemetriou:
@vmarkovtsev Hey Vadim, could we get this merged please?

vmarkovtsev (approved):
The change is backward compatible, we should be good :+1:

PR #246, opened 2020-09-22
Make fetch_val call fetch_one for type conversion

se7entyse7en:
Since `self._connection.fetch_val` do not perform type conversion, `fetch_val` is changed to call `fetch_one`.


vmarkovtsev (approved):
Thanks for such a great bug fix! Very nice code.

vmarkovtsev:
Relates to https://github.com/encode/databases/pull/132

Issue #247, opened 2020-09-28
Unsafe DB password will fail urlsplit in DatabaseURL


librehat:
```python
url = 'postgres://username:[password@localhost:80/dbname'
DatabaseURL(url)  # an exception will be raised here from `urlsplit`
```

The root cause is the use of `urllib.parse` and the assumption that all components can be represented in the URL _safe_ unquoted. As you can see, that is not always the case in real life (one of our servers has generated a password that has a `[` but no `]`)

librehat:
One simple patch I can think of right now is to treat all components potentially _quoted_, in which case the `urlsplit` method continues to work, but each component (username, password etc.) should be `unquote` in `DatabaseURL`. This change _shouldn't_ break anything (unless % escaped characters are in fact expected password).

A more comprehensive change is to use keyword parameters to specify each component (username, password, etc.) instead of parsing a URL.

tomchristie:
Cross-linking to the convo on gitter... https://gitter.im/encode/community?at=5f720952417bf140aa1b47e5

librehat:
`URL` implementation in `httpx` as a reference: https://github.com/encode/httpx/blob/32d37cfdf1697f0dbacbc2233ee34606369be87b/httpx/_models.py#L63

librehat:
Do we know vaguely when we will see a new release of `databases`? 👀

vmarkovtsev:
I am working on this right now. I wanted to merge as many PRs as possible before issuing a release.

vmarkovtsev:
https://github.com/encode/databases/pull/255

PR #248, opened 2020-09-29
Unquote username and password in DatabaseURL #247

librehat:


databases/core.py
```diff
@@ -411,13 +416,26 @@ def driver(self) -> str:
             return ""
         return self.components.scheme.split("+", 1)[1]
 
+    @property
+    def userinfo(self) -> typing.Optional[bytes]:
+        if self.components.username:
+            info = self.components.username
+            if self.components.password:
+                info += ":" + self.components.password
+            return info.encode("ascii")
```
vmarkovtsev, line 425: Some DBs, for example, Postgres, allow using any crazy Unicode user names and passwords.

librehat: can change this to "utf-8", i chose `ascii` because it was how it is encoded into bytes in `httpx`

vmarkovtsev (approved):

vmarkovtsev:
No idea why httpx encodes `userinfo` to `bytes`, Anyway, we can always change to `str`.

PR #250, opened 2020-10-08
Allow password as a callable when using Postgres backend

nwillems:
This change is intended for a case where you have changing passwords for the database, eg when using short-lived tokens.

Specifically, this allows one to use databases together with RDS IAM access, where a token is generated for new connections.

### DIFF
--- a/databases/backends/postgres.py
+++ b/databases/backends/postgres.py
@@ -44,7 +44,7 @@ class PostgresBackend(DatabaseBackend):
     def _get_connection_kwargs(self) -> dict:
         url_options = self._database_url.options
 
-        kwargs = {}
+        kwargs = {}  # type: typing.Dict[str, typing.Any]
         min_size = url_options.get("min_size")
         max_size = url_options.get("max_size")
         ssl = url_options.get("ssl")
--- a/databases/backends/postgres.py
+++ b/databases/backends/postgres.py
@@ -62,15 +62,15 @@ class PostgresBackend(DatabaseBackend):
 
     async def connect(self) -> None:
         assert self._pool is None, "DatabaseBackend is already running"
-        kwargs = self._get_connection_kwargs()
-        self._pool = await asyncpg.create_pool(
+        kwargs = dict(
             host=self._database_url.hostname,
             port=self._database_url.port,
             user=self._database_url.username,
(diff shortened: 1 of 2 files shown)


nwillems:
I'm at a loss here - not quite sure how to get to those 100% coverage. Any help would be greatly appreciated :-)

nwillems:
> 
> 
> I'm at a loss here - not quite sure how to get to those 100% coverage. Any help would be greatly appreciated :-)

Found out that it was coverage of the supplied function that was missing.

Typing is now also added.

Should be golden :-)

databases/backends/postgres.py
```diff
@@ -67,7 +71,6 @@ async def connect(self) -> None:
             host=self._database_url.hostname,
             port=self._database_url.port,
             user=self._database_url.username,
-            password=self._database_url.password,
```
vmarkovtsev, line 70: I'd rather do it with fewer changes. We can get away with the following:
1. Revert everything.
2.
```python
kwargs = dict(
    host=self._database_url.hostname,
    port=self._database_url.port,
    user=self._database_url.username,
    password=self._database_url.password,
    database=self._database_url.database,
)
kwargs.update(self._get_connection_kwargs())
self._pool = await asyncpg.create_pool(**kwargs)
```
You can then pass `password` to `options / self._options`.

vmarkovtsev (changes requested):
Thanks for your contribution!

databases/backends/postgres.py
nwillems: I agree. Like the updated code?

nwillems:
@vmarkovtsev I've updated the code and it should be good - thanks for the review. Hope it fits :-)

vmarkovtsev:
Great, thanks for the ping :+1:

vmarkovtsev (approved):
Glad that it worked!

ugtar:
looks like the fix for this actually also fixes a regression that makes it impossible to connect to a postgres db over a unix socket (#257 ). The PR I wrote for that is nearly identical to this. I guess I can close that issue and the associated PR.

vmarkovtsev:
Thank you @ugtar 👍

PR #251, opened 2020-10-10
Issue 249 postponed configuration

### LINKED ISSUE #249: Postponed configuration of database
Session from sqlalchemy [allows](https://docs.sqlalchemy.org/en/13/orm/session_api.html#sqlalchemy.orm.session.sessionmaker.configure) postponed configuration:
```python
SessionFactory = sessionmaker()
SessionFactory.configure(bind=create_engine('sqlite://'))

session_instance = SessionFactory()
```
Maybe `Database` should be also has `configure` method? This allows to create `Database` object at import time but set parameters for connection (url, pool size, etc) at runtime.

Right now I use FastAPI framework with factory pattern:
```python
def create_app() -> FastAPI:
    app = FastAPI()

    cfg = app.state.cfg = config.parse_config()

    database = app.state.database = create_database(cfg.database)
```
This way I have database instance when handling requests. But I cannot set decorators on function at import time like this because I dont have database instance at import time because I parse config at runtime:
```python
@database.transaction()  # cannot do this
async def create_users(request):
    ...
```
So possible API can be look like this:
```python
database = Database()


def create_app() -> FastAPI:
    app = FastAPI()

    cfg = app.state.cfg = config.parse_config()

    database.initialize(url=cfg.url)

    @app.on_event("startup")
    async def startup():
        await database.connect()

    app.include_route(create_users)


@database.transaction()
async def create_users(request):
    database.execute(create_query)
    return database.fetch_one(query)
```
Also should note, that `configure` method should allow multiple calls, so in tests fixture with `function` scope can recreate ASGI app multiple times (call factory method), which in turn calls `configure`.

artslob:
Closes #249

### DIFF
--- a/databases/core.py
+++ b/databases/core.py
@@ -103,10 +103,15 @@ class Database:
         assert issubclass(backend_cls, DatabaseBackend)
         self._backend = backend_cls(self.url, **self.options)
 
+    def _should_be_inited(self):
+        if not self._inited:
+            raise ValueError("Databases object is not configured")
+
     async def connect(self) -> None:
         """
         Establish the connection pool.
         """
+        self._should_be_inited()
         assert not self.is_connected, "Already connected."
 
         await self._backend.connect()
@@ -130,6 +135,7 @@ class Database:
         """
         Close all connections in the connection pool.
         """
+        self._should_be_inited()
         assert self.is_connected, "Already disconnected."
 
         if self._force_rollback:
@@ -202,6 +208,8 @@ class Database:
                 yield record
 
     def connection(self) -> "Connection":
+        self._should_be_inited()
+
         if self._global_connection is not None:
             return self._global_connection
 
--- a/databases/core.py
+++ b/databases/core.py
@@ -63,6 +63,7 @@ class Database:
         self.is_connected = False
 
         self._force_rollback = False
+        self._inside_force_rollback_context = False
 
         self._backend = None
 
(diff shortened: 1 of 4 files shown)


vmarkovtsev (commented):
This is a big change that touches some of the most critical code and complicates the object lifecycle considerably. I am not even sure we want the deferred configuration in the first place.

artslob:
Yeah, PR and issue are open for discussion to determine whether this PR will be useful. I just think that sometimes its not convenient nor beautiful to parse config at import time. I prefer factory pattern so made this PR.

rafalp:
IMHO there's problem with our API design, but band-aiding it will not help, only make the problem harder to deal with later.

`Core` object wraps the process of strategy creation, but there's no way to extend this process without modifying internals themselves. This bites not only people who need to configure backend-specific extra options not supported by the core, but also those interested in developing support for other databases.

artslob:
Can I do something to improve this PR? Or should I close it then?

artslob:
Maybe create some proxy/wrapper around `Database` class? This way `Database` would not be changed:
```python
class DatabaseWrapper:
    def __init__(self):
        self.db = None

    def configure(self, **all_rapams):
        self.db = Database(**all_rapams)

    def check_configured(self):
        if not self.db:
            raise ValueError

    def connect(self):
        # proxies all methods to inner object
        self.check_configured()
        return self.db.connect()


class Database:
    # this class would not be changed
    pass


db = DatabaseWrapper()


def main():
    params = parse_env_params()
    db.configure(**params)
    connect = db.connect()
```

PR #256, opened 2020-10-21
Remove psycopg2 and pymysql from the direct dependencies

vmarkovtsev:
Related to https://github.com/encode/databases/pull/204

This should fix my own builds and I guess other people's too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## 0.4.1 (October 21st, 2020)
+
+### Fixed
+
+* Remove package dependency on the synchronous DB drivers (#256)
+
 ## 0.4.0 (October 20th, 2020)
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ $ pip install databases[sqlite]
 ```
 
 Driver support is provided using one of [asyncpg][asyncpg], [aiomysql][aiomysql], or [aiosqlite][aiosqlite].
+Note that if you are using any synchronous SQLAlchemy functions such as `engine.create_all()` or [alembic][alembic] migrations then you still have to install a synchronous DB driver: [psycopg2][psycopg2] for PostgreSQL and [pymysql][pymysql] for MySQL.
 
 ---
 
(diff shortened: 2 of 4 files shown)


vmarkovtsev:
cc @retnikt

daviskirk:
Haven't upgraded but I know that the binary/non-binary problem has gotten me before, especially when trying to explain to others what they have to do to not get errors when installing stuff.  So just having everyone decide what they want as suggested here is probably better.

subdavis:
Just to be clear, this broke build for everyone who uses `databases[postgresql]` and `psycopg2-binary`, right?

I'm surprised there isn't more voting/commenting on this thread.

For anyone playing along at home, `"databases[postgresql]<0.4.0",` should temporarily fix your build errors.

vmarkovtsev:
Version pinning ftw, yes. There is another option:
```
databases==0.4.0
asyncpg==0.21.0
```

polyrand:
Yes, this broke my builds too. I think the option given by Tom: https://github.com/encode/databases/pull/204#issuecomment-622333939 is very reasonable. Thanks for the fast response and the work done on this!

vmarkovtsev:
@encode/maintainers can please somebody approve this deal breaking fix.

rsommerard:
I'm having this issue too, @encode/maintainers please consider this PR.

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## 0.4.1 (October 21st, 2020)
+
+### Fixed
+
+* Remove package dependency on the synchronous DB drivers (#256)
+
```
florimondmanca, line 12: Are we okay with bumping to `0.5.0`? I think this qualifies as an entry for a "Changed" section, rather than "Fixed". That way people that are currently pinned to `0.4.*` will be fine.

I'm not sure I understood the broken build issue with these direct dependencies, though, but I think sticking closer to SemVer is a good idea here?

florimondmanca (approved):
Sounds good — let's consider a `0.5.0` bump perhaps. (Or, alternatively, let's skip bumping versions in this PR, and issue a proper release PR later when we're ready to issue `0.5.0`.)

CHANGELOG.md
vmarkovtsev: Besides, we have to bump the version, otherwise it does deploy to pypi and many people continue suffering.

Thanks for the approval, I will carry on with the release now.

vmarkovtsev:
https://github.com/encode/databases/releases/tag/0.4.1

CHANGELOG.md
florimondmanca: Makes sense!

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## 0.4.1 (October 21st, 2020)
```
florimondmanca, line 7: Btw, the release date is a bit out of sync now. :-)

Should be:

```
## 0.4.1 (November 16th, 2020)
```

Issue #72, opened 2019-03-22
Support for "default" parameter in sqlalchemy.Column


dgilland:
Whenever I have SQLAlchemy table column with a default value (e.g. `Column("verified", Boolean(), default=False)`), the resulting insert statement generated by `databases` sets a `null` value for the column when that column isn't provided (e.g. `db.execute(mytable.insert(), {...})` where "verified" isn't set).

Are there plans to support column defaults in insert statements?

tomchristie:
Good question.
Not necessarily, no. We might consider it out of scope. It’s a bit of a blurry line, since everything else in the column data is all about the actual database schema, but default is an application-level indicator. We might choose not to support it here, but only do so at other layers of abstraction eg. The ORM package does have default support.

Not sure.

dgilland:
For me personally, I'm not interested in using an ORM so it would be nice if application-level defaults were supported. However, I completely understand if you think that's something out-of-scope for `databases`.

gvbgduh:
@dgilland you might consider `server_default` which will apply the default at the DB level, like `now()` func.
Another thing to consider at the app level is to delegate this to some validation schema/class/model.
As example, if data for insertion comes from the payload you might want to validate it before actually committing it to the db. One of many ways to do it might be a `marshmallow` schema with some `@post_load` action.
I am quite sure [typesystem](https://github.com/encode/typesystem) can do similar stuff as well.

dgilland:
@gvbgduh Thanks for the suggestions! :+1:

euri10:
@gvbgduh what's the `server_default` you're referring to, I greped the source and found nothing?

tomchristie:
@euri10 Search in SQLAlchemy, not this project - it’s a table definitions thing.

ryananguiano:
You can generate the defaults yourself like this:
```python
def get_defaults(db_table):
    defaults = {}
    for column in db_table.columns:
        if column.default is not None:
            value = column.default.arg
            if callable(value):
                value = value()
            defaults[column.name] = value
    return defaults
```

jordic:
In my latest relase of asyncom I support this case for simple cases (values and callables.) there are also other defaults supported.

https://github.com/vinissimus/asyncom/blob/master/asyncom/om.py#L167

I just seen that I had still not committed the tests.

mivade:
[...]

PR #266, opened 2020-11-12
Fix sqlalchemy default values for insert and update queries

ckkz-it:
This is what actually `aiopg` does https://github.com/aio-libs/aiopg/blob/master/aiopg/sa/engine.py#L17#L30

Fixes https://github.com/encode/databases/issues/72
I think this solution is less tricky than https://github.com/encode/databases/pull/206/

### DIFF
--- a/databases/backends/postgres.py
+++ b/databases/backends/postgres.py
@@ -20,7 +20,9 @@ logger = logging.getLogger("databases")
 _result_processors = {}  # type: dict
 
 
-class APGCompiler_psycopg2(ConstructDefaultParamsMixin, pypostgresql.dialect.statement_compiler):
+class APGCompiler_psycopg2(
+    ConstructDefaultParamsMixin, pypostgresql.dialect.statement_compiler
+):
     pass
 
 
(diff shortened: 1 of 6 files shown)


tests/test_databases.py
```diff
@@ -917,6 +1009,7 @@ async def run_database_queries():
         async with database:
 
             async def db_lookup():
+
```
astret-soft, line 1012: left \n

databases/backends/postgres.py
```diff
@@ -10,11 +10,20 @@
...
+from databases.backends.common import ConstructDefaultParamsMixin
 from databases.core import LOG_EXTRA, DatabaseURL
 from databases.interfaces import ConnectionBackend, DatabaseBackend, TransactionBackend
 
 logger = logging.getLogger("databases")
 
+_result_processors = {}  # type: dict
+
+
+class APGCompiler_psycopg2(
```
acnebs, line 22: Seems strange to call this `APGCompiler_psycopg2`, given that the db backend used for postgres in databases is `asyncpg`, not `psycopg2`.

ckkz-it: this is how aiopg calls it
https://github.com/encode/databases/blob/master/databases/backends/aiopg.py#L32#L35

acnebs: Right but that's because aiopg is just a wrapper around psycopg2, asyncpg is not.

PR #273, opened 2020-12-10
Fix memory leak with asyncpg, for SQLAlchemy generic functions

### LINKED ISSUE #272: Memory leak with asyncpg, for SQLAlchemy generic functions
## Summary

The `postgres` backend includes a cache for result processors, which makes the library to leak memory when some specific SQLAlchemy generic functions are used.

## Description

The `_result_processors` dictionary is used as a global cache: https://github.com/encode/databases/blob/8cd551436cf39f3db2913c5b3338a9e89177af41/databases/backends/postgres.py#L19

Every time a record is accessed, the `Record.__getitem__` method is called, and that's where the cache is filled, considering the data type of the accessed field (e.g. `Integer()`, `Date()`...). Let's take into account that the keys for this cache are instances for any of these SQLAlchemy types. https://github.com/encode/databases/blob/8cd551436cf39f3db2913c5b3338a9e89177af41/databases/backends/postgres.py#L123-L127

So, for this cache to be successfully used, we need type instances that are reused during the application lifetime. That's how SQLAlchemy works, when defining tables and their columns.

**However**, SQLAlchemy also provides other SQL functions that don't work in this same way. Specifically, any function that **inherits from [`GenericFunction`](https://github.com/sqlalchemy/sqlalchemy/blob/6eda3aa8c5f58e415a49ac8dc73eb8608f4be306/lib/sqlalchemy/sql/functions.py#L696)** but **does not inherit from [`ReturnTypeFromArgs`](https://github.com/sqlalchemy/sqlalchemy/blob/6eda3aa8c5f58e415a49ac8dc73eb8608f4be306/lib/sqlalchemy/sql/functions.py#L835)**.

This is because the `ReturnTypeFromArgs` constructor reuses the type of the column it wraps: https://github.com/sqlalchemy/sqlalchemy/blob/6eda3aa8c5f58e415a49ac8dc73eb8608f4be306/lib/sqlalchemy/sql/functions.py#L840

But the `GenericFunction` constructor initializes a new data type every time it's called (unless an instantiated type is provided, something not common in this scenario: https://github.com/sqlalchemy/sqlalchemy/blob/6eda3aa8c5f58e415a49ac8dc73eb8608f4be306/lib/sqlalchemy/sql/functions.py#L796-L798

As a new type is initialized every time, it's added to the `_result_processors` dictionary, which keeps growing and leaking memory, with entries that won't be reused for these particular type instances.

## Test case

This is a simple test case I built to reproduce the issue. It includes functions that leak memory as part of the `test_functions_with_memory_leak` test case parametrization, and functions that do not leak memory (because of being subclasses of `ReturnTypeFromArgs`) as part of the `test_functions_without_memory_leak` test case parametrization.

Requirements:
```
databases[postgresql]==0.4.1
pytest
pytest-asyncio
```

Test cases:
```
import asyncio

from databases.backends.postgres import _result_processors
import pytest
from sqlalchemy import column, func, select, table


SCORES_TABLE = table('HighScores', column('id'), column('name'), column('score'))


@pytest.fixture(scope='session')
def event_loop():
    """Needed to have Session-scoped fixtures."""
    return asyncio.new_event_loop()


@pytest.fixture(scope='session', autouse=True)
async def db():
    """Create a database instance, and connect to it."""
    database = Database('postgresql://postgres:postgres@localhost/db')
    await database.connect()
    await database.execute(
        query="""
            CREATE TABLE IF NOT EXISTS "HighScores"
            (id INTEGER PRIMARY KEY, name VARCHAR(100), score INTEGER)
        """,
    )
    await database.execute(
        query="""
            INSERT INTO "HighScores" (id, name, score)
            VALUES (1, 'TEST', 1000)
            ON CONFLICT DO NOTHING
        """,
    )
    yield database
    await database.disconnect()


@pytest.fixture(autouse=True)
async def clean_result_processors():
    """After each test, clean the `_result_processors` internal dictionary."""
    yield
    _result_processors.clear()


@pytest.mark.asyncio
@pytest.mark.parametrize('query_func', (
    lambda: select([func.count()]).select_from(SCORES_TABLE),
    lambda: select([func.concat(SCORES_TABLE.c.name, SCORES_TABLE.c.name)]).select_from(SCORES_TABLE),
    lambda: select([func.current_date()]),
    lambda: select([func.now()]),
))
async def test_functions_with_memory_leak(query_func, db):
    """This test is parametrized with some of the SQLAlchemy functions that generate the memory leak."""
    for _ in range(100):
        query = query_func()
        record = await db.fetch_one(query=query)
        # Access record fields, to trigger Record.__getitem__
        [record[item] for item in record._row.keys()]
        assert len(_result_processors) == len(record._row)


@pytest.mark.asyncio
@pytest.mark.parametrize('query_func', (
    lambda: select([func.sum(SCORES_TABLE.c.score)]).select_from(SCORES_TABLE),
    lambda: select([func.min(SCORES_TABLE.c.score)]).select_from(SCORES_TABLE),
    lambda: select([func.max(SCORES_TABLE.c.score)]).select_from(SCORES_TABLE),
    lambda: select([func.coalesce(SCORES_TABLE.c.score, 0)]).select_from(SCORES_TABLE),
))
async def test_functions_without_memory_leak(query_func, db):
    """This test is parametrized with some of the SQLAlchemy functions that do not generate the memory leak."""
    for _ in range(100):
        query = query_func()
        record = await db.fetch_one(query=query)
        # Access record fields, to trigger Record.__getitem__
        [record[item] for item in record._row.keys()]
        assert len(_result_processors) == len(record._row)
```

adamantike:
Use SQLAlchemy caching logic for data type's result processors, instead of handling it in this library.

Fixes #272

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## Unreleased
+
+### Fixed
+
+* Fix memory leak with asyncpg, for SQLAlchemy generic functions (#273)
+
 ## 0.4.1 (November 16th, 2020)
 
 ### Fixed
--- a/databases/backends/postgres.py
+++ b/databases/backends/postgres.py
@@ -16,9 +16,6 @@ from databases.interfaces import ConnectionBackend, DatabaseBackend, Transaction
 logger = logging.getLogger("databases")
 
 
-_result_processors = {}  # type: dict
-
-
 class PostgresBackend(DatabaseBackend):
     def __init__(
         self, database_url: typing.Union[DatabaseURL, str], **options: typing.Any
@@ -120,11 +117,7 @@ class Record(Mapping):
         else:
             idx, datatype = self._column_map[key]
         raw = self._row[idx]
-        try:
-            processor = _result_processors[datatype]
-        except KeyError:
-            processor = datatype.result_processor(self._dialect, None)
-            _result_processors[datatype] = processor
+        processor = datatype._cached_result_processor(self._dialect, None)
 
         if processor is not None:
             return processor(raw)


vmarkovtsev (approved):

adamantike:
@vmarkovtsev, if there are no planned changes in the short term, do you mind tagging and releasing a new patch version, to have this fix available for current clients? :)

adamantike:
@vmarkovtsev gentle ping!

rafalp:
Looks like this should ship as v0.4.2. I'll look around how release process looks for Encode repos and see if I can do this.

rafalp:
@adamantike this fix has been released as 0.4.2

adamantike:
Great, thank you @rafalp!

PR #283, opened 2021-01-29
Fixed counter on failed connection

qweryty:
Sometimes `self._connection.acquire()` can throw an exception and `self._connection_counter` will remain 1, in that case, if we try to reconnect `__aenter__` will always skip `self._connection.acquire()` and we won't be able to make any further requests to database.

### DIFF
--- a/databases/core.py
+++ b/databases/core.py
@@ -215,8 +215,12 @@ class Connection:
     async def __aenter__(self) -> "Connection":
         async with self._connection_lock:
             self._connection_counter += 1
-            if self._connection_counter == 1:
-                await self._connection.acquire()
+            try:
+                if self._connection_counter == 1:
+                    await self._connection.acquire()
+            except Exception as e:
+                self._connection_counter -= 1
+                raise e
         return self
 
     async def __aexit__(


taybin:
I believe I have been encountering this issue, so it would be great to see this or something similar to it merged.

taybin:
Looks like it  just needs a unittest.

databases/core.py
```diff
@@ -237,8 +237,12 @@ def __init__(self, backend: DatabaseBackend) -> None:
     async def __aenter__(self) -> "Connection":
         async with self._connection_lock:
             self._connection_counter += 1
```
aminalaee, line 239:
```suggestion
            if self._connection_counter == 0:
                await self._connection.acquire()
                self._connection_counter += 1
```

aminalaee (commented):
Maybe this can be simplified a bit?

databases/core.py
aminalaee: If any exception happens it won't be incremented at all, so we don't have to decrement in the except part.

aminalaee: @taybin @qwerty what do you think? This is simpler and won't need the extra test,

taybin: Does this still work with concurrent requests?  It removes the semaphore.  In which case it doesn't even need the counter.  The connection lock would be enough.

But is the connection lock enough or do we need the semaphore?

aminalaee: @taybin I'm not really sure, I'll try a test, if it doesn't work out, we'll merge it as is.

aminalaee:
I'll try to add the test for new code, @taybin feel free to add that if you like.

taybin:
I'm working on a unittest for this right now.

taybin:
@aminalaee I have a PR with @qweryty's code and a unittest at #385

PR #295, opened 2021-03-08
Fix DatabaseURL string representation

KonstantinTogoi:
This fixes string representation of `DatabaseURL` objects.

Here is a script to reproduce problem:

```python
from databases import Database, DatabaseURL
from pydantic import BaseSettings, PostgresDsn


class Settings(BaseSettings):
    dsn: PostgresDsn
 

dsn = DatabaseURL(Settings(dsn='postgres://user:pass@localhost/name').dsn)
db = Database(dsn)
str(db.url)
# TypeError: __init__() needs keyword-only argument scheme
```

The problem appears when `DatabaseURL`'s argument `url` is an instance of a class that derived from `str` and has overriden `__repr__()` method:

```python
from databases import DatabaseURL

Str = type('Str', (str, ), {'__repr__': lambda self: 'Lorem ipsum dolor sit amet...'})

dsn = DatabaseURL(Str('postgres://user:pass@localhost/name'))
dsn
# DatabaseURL('postgres://user:********@localhost/name')
print(dsn)
# postgres://user:pass@localhost/name
str(dsn)
# Lorem ipsum dolor sit amet...
```

After this fix:

```python
from databases import DatabaseURL

Str = type('Str', (str, ), {'__repr__': lambda self: 'Lorem ipsum dolor sit amet...'})

dsn = DatabaseURL(Str('postgres://user:pass@localhost/name'))
dsn
# DatabaseURL('postgres://user:********@localhost/name')
print(dsn)
# postgres://user:pass@localhost/name
str(dsn)
# 'postgres://user:pass@localhost/name'
```

### DIFF
--- a/databases/core.py
+++ b/databases/core.py
@@ -392,7 +392,7 @@ class DatabaseURL:
         if isinstance(url, DatabaseURL):
             self._url: str = url._url
         elif isinstance(url, str):
-            self._url = url
+            self._url = str(url)
         else:
             raise TypeError(
                 f"Invalid type for DatabaseURL. Expected str or DatabaseURL, got {type(url)}"
--- a/tests/test_database_url.py
+++ b/tests/test_database_url.py
@@ -17,6 +17,22 @@ def test_database_url_repr():
     assert repr(u) == "DatabaseURL('postgresql://username:********@localhost/name')"
 
 
+def test_database_url_str():
+    Str = type("Str", (str,), {"__repr__": lambda self: ""})
+
+    u = DatabaseURL(Str("postgresql://localhost/name"))
+    assert str(u) == "postgresql://localhost/name"
+
+    u = DatabaseURL(Str("postgresql://username@localhost/name"))
+    assert str(u) == "postgresql://username@localhost/name"
+
+    u = DatabaseURL(Str("postgresql://username:password@localhost/name"))
+    assert str(u) == "postgresql://username:password@localhost/name"
+
+    u = DatabaseURL(Str(f"postgresql://username:{quote('[password')}@localhost/name"))
+    assert str(u) == "postgresql://username:%5Bpassword@localhost/name"
+
+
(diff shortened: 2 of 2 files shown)


rafalp:
Seems odd. Is this fix for case when somebody uses custom str type to prevent DB credentials from accidentally printing somewhere?

KonstantinTogoi:
> Seems odd. Is this fix for case when somebody uses custom str type to prevent DB credentials from accidentally printing somewhere?

If somebody uses custom str type to prevent credentials from accidentally printing then the credentials will not be printed.

I checked with the following example:

```python
from databases import DatabaseURL


class SecretStr(str):
    def __new__(cls, public, **kwargs):
        arg = cls.build(**kwargs) if public is None else public
        return str.__new__(cls, arg)

    def __init__(self, public, *, prefix='', secret='', suffix=''):
        str.__init__(public)
        self.public = public
        self.prefix = prefix
        self.secret = secret
        self.suffix = suffix

    def __repr__(self):
        return self.public or self.prefix + self.obscure_secret + self.suffix

    def __str__(self):
        return self.public or self.prefix + self.obscure_secret + self.suffix

    @classmethod
    def build(cls, *, prefix='', secret='', suffix=''):
        return prefix + secret + suffix

    @property
    def obscure_secret(self):
        return '******'


secret = SecretStr(None, prefix='[public prefix]', secret='secret', suffix='[public suffix]')
public = SecretStr('str without secret part')

secret_dsn = DatabaseURL(secret)
public_dsn = DatabaseURL(public)

print(repr(secret_dsn))
print(str(secret_dsn))
print(repr(public_dsn))
print(str(public_dsn))
```

Output:

```
DatabaseURL([public prefix]******[public suffix])
[public prefix]******[public suffix]
DatabaseURL('str without secret part')
str without secret part
```

aminalaee:
Closing this as out of date.

PR #299, opened 2021-03-15
fix: support sqlalchemy 1.4

PrettyWood:


databases/backends/postgres.py
```diff
@@ -104,8 +104,9 @@ def __init__(
         self._dialect = dialect
         self._column_map, self._column_map_int, self._column_map_full = column_maps
 
-    def values(self) -> typing.ValuesView:
-        return self._row.values()
+    @property
+    def _mapping(self) -> asyncpg.Record:
+        return self._row
```
PrettyWood, line 109: I would like your insight on this matter.

For `postgres`, `row.values()` was still working thanks to this `Record` class but for all other backends we now need to do `row._mapping.values()` with sqlalchemy >= 1.4

Do we want to:
1) keep `values()` only for `postgres`
2) keep `values()` for all backends and we can attach the deprecated `property` to the `sqlalchemy.Row` class
3) let it like this and mimic sqlalchemy API also for `postgres`

Thanks for this very nice lib!

ToGoBananas:
@vmarkovtsev please join

uSpike:
`setup.py` needs to be updated to specify `install_requires=["sqlalchemy>=1.4,<2", ...` or a similar constraint.

tomchristie:
New release 0.4.3 to properly pin SQLAlchemy until this PR is resolved. https://pypi.org/project/databases/

(This PR should update the pin to `<1.5`, so we don't get bitten again.)

PrettyWood:
@tomchristie do we want to support both 1.3 and 1.4 or 1.4 only?

tomchristie:
Ideally both, at least for a time. (But if that was really awkward I guess we could consider just 1.4, with some clear notes for users on which older version of `databases` to use for 1.3 etc)

PrettyWood:
@tomchristie Ok no problem. Is it ok if I update CI pipeline to run all tests with both versions?

Euromance:
I guess dropping support for 1.3 is okay since `pip` 's new dependency resolver would automatically download an older version of `databases` if the user's environment requires `SQLAlchemy 1.3`.

An ugly solution to support both 1.3 and 1.4 can be seen [here in SQLAlchemy-Utils 0.37](https://github.com/kvesteri/sqlalchemy-utils/compare/0.36.8...0.37.0)

tomchristie:
@Euromance Fair point. I'm okay with requiring 1.4+ then.

databases/backends/postgres.py
Euromance: imo It'd be better to keep `row.values()` for all backends since looking at test changes I can see that `row.keys()` is still present so `row._mapping.values()` would bring an inconsistency

PrettyWood:
Cool! I'll get back on it next Thursday when I have more time

databases/backends/postgres.py
PrettyWood: Hmmm I don't really want to add `row.values()` for all backends since it would need some monkeypatching or wrapper around the SQLAlchemy objects.
I reckon the idea of `databases` is to keep things simple and to stay close to SQLAlchemy when possible.
This is why I went with updating postgres backend only to mimic SQLAlchemy behaviour

Euromance: My bad, I started my comment with keep since that inconsistency was triggering me, although after reading SQLAlchemy's changelogs and docs I immediately changed my mind about keeping `values` but did not change the original paragraph. 

I do agree with decision to keep the API as close to SQLAlchemy's as possible.

Just that test having both `row.values` and `row._mapping.keys` misled me.

Euromance (commented):
I guess you could also test that deprecated methods throw their warnings

PrettyWood:
@Euromance Yes it was planned thanks 👍 I first wanted to have your feedback on the interface. Glad we agree! Coverage should be good now

nikolayanguelov:
Do you guys have any update on the ETA on when this might get merged and released to the public?

Thank you!

pvicente:
Hi, is there any plan to merge this PR? When are you planning to do a release with sqlalchemy >= 1.4 support? 🙏🏼

ToGoBananas (approved):

tomchristie:
@PrettyWood If you'd like to get this in then let me know and can add you to the maintainers team, and walk you through the release process.

PrettyWood:
With pleasure @tomchristie!

tomchristie:
@PrettyWood - Okay, I've sent you an invite. I'll leave getting this reviewed/merged/conflict resolved/whatever up to you - tho let us know if you need anything. Then once that's done let's walk through how you can roll a release.

PrettyWood:
@tomchristie I rebased on `master`. I'll wait for a review from a member. If everything is fine, I feel like the release process is straightforward (PR to bump version with changelog update + tag that triggers github worfklow to publish on pypi). Am I missing something?

setup.py
```diff
@@ -48,7 +48,7 @@ def get_packages(package):
     packages=get_packages("databases"),
     package_data={"databases": ["py.typed"]},
     data_files=[("", ["LICENSE.md"])],
-    install_requires=['sqlalchemy<1.4', 'aiocontextvars;python_version<"3.7"'],
+    install_requires=['sqlalchemy>=1.4,<1.5', 'aiocontextvars;python_version<"3.7"'],
```
aminalaee, line 51: As far as I understand your change is backwards compatible with older versions of `sqlalchemy`. Do we need to pin this to `>= 1.4`?

PrettyWood: No it's not. It will break for example when importing `sqlalchemy.engine.cursor` (see the related issue)
I asked above if I should make it compatible (in which case I would have done a util module that imports the right methods depending on the version) but [we agreed](https://github.com/encode/databases/pull/299#issuecomment-832555219) it was not necessary. If people want to use sqlachemy 1.3, pip will install `databases 0.4.3`

aminalaee: Sorry, I missed that.

aminalaee (approved):

sometato:
Is this ready to be merged? Thank you!

Feijo:
I am also waiting for this release

aminalaee:
@PrettyWood Do you want to make any other changes or are you waiting for more reviewers?

Issue #327, opened 2021-04-27
Async execution of transactions results in deadlock 


goteguru:
**Concurrent execution of transactions deadlocks.**

Minimal example:

```
async def test_task(db): 
     async with db.transaction():
            await db.fetch_one("SELECT 1")

async with Database(database_url) as database:
     await database.fetch_one("SELECT 1")
     tasks = [test_task(database) for i in range(4)]
     await asyncio.gather(*tasks)
```
This is especially troublesome, because if we remove the 6th line (or move it to the end of the function) everything works ok. Very confusing for the user. 

Please notice it's a very common use-case if somebody is using an async library. The user might want to send multiple mails in parallel (and record the result) or pull remote resources asynchronously and do something in the database. 

**Detailed analysis:**

Connections are stateful in most (probably all) database engines. _Databases_ try to hide the connection by use of contexvars, which is created per Task. The Database object creates the Connection on the fly if not exits. Unfortunately if we do anything before starting a transaction, the connection will be bound and all subsequent transactions will be executed using that connection compromising the state if executed concurrently. 

This is not a problem if the connections are hierarchical. The library has a transaction buffer to trace this. Unfortunately this buffer causes trouble if the transactions are executed parallel in different Tasks.  If task "A" started first, then task "B", but task "A" finishes before task "B" (which is possible) the transaction boundaries will overlap and the module deadlocks with exception (sqlite) or unhandled (postgres). 

Possible workaround is to wrap everything in Tasks like:
```
...
async with Database(database_url) as database:
     await asyncio.create_task(database.fetch_one("SELECT 1"))
     tasks = [test_task(database) for i in range(4)]
     await asyncio.gather(*tasks)
```
However it is slightly counter-intuitive and even useless if we already executed anything before several levels higher in the callstack (ie. having a connection). 
 
One intuitive solution could be copying the context for every new root transactions (ie. creating new connection for every *root* transaction). See the proposed fix in the pull request #328 .

PR #328, opened 2021-04-27
fix concurrency issues of parallel transactions (#327)

goteguru:
This is the proposed solution for async transaction execution failure (see #327 )  

Tests are running fine for postgres, mysql, sqlite. 

I had to change the test tests/test_databases.py because it incorrectly expects  database.transaction() to use the same connection as the connection context around it, which is certainly not true (and actually the part of the problem). If I acquire a connection, I have to use that connection for a transaction. (Requesting the database directly may give any other connection from the pool). 

All comments are welcome.

### DIFF
--- a/databases/core.py
+++ b/databases/core.py
@@ -15,8 +15,10 @@ from databases.interfaces import ConnectionBackend, DatabaseBackend, Transaction
 
 if sys.version_info >= (3, 7):  # pragma: no cover
     from contextvars import ContextVar
+    import contextvars as contextvars
 else:  # pragma: no cover
     from aiocontextvars import ContextVar
+    import aiocontextvars as contextvars
 
 try:  # pragma: no cover
     import click
@@ -173,6 +175,11 @@ class Database:
             async for record in connection.iterate(query, values):
                 yield record
 
+    def _new_connection(self) -> "Connection":
+        connection = Connection(self._backend)
+        self._connection_context.set(connection)
+        return connection
+
     def connection(self) -> "Connection":
         if self._global_connection is not None:
             return self._global_connection
@@ -180,14 +187,22 @@ class Database:
         try:
             return self._connection_context.get()
         except LookupError:
-            connection = Connection(self._backend)
-            self._connection_context.set(connection)
-            return connection
+            return self._new_connection()
 
     def transaction(
         self, *, force_rollback: bool = False, **kwargs: typing.Any
     ) -> "Transaction":
-        return Transaction(self.connection, force_rollback=force_rollback, **kwargs)
+        try: 
+            connection = self._connection_context.get()
+            if not connection._transaction_stack:
+                newcontext = contextvars.copy_context()
+                get_conn  = lambda: newcontext.run(self._new_connection)
+            else:
+                get_conn = self.connection
+        except LookupError:
+            get_conn = self.connection
+
+        return Transaction(get_conn, force_rollback=force_rollback, **kwargs)
 
     @contextlib.contextmanager
     def force_rollback(self) -> typing.Iterator[None]:
@@ -357,6 +372,7 @@ class Transaction:
     async def start(self) -> "Transaction":
         self._connection = self._connection_callable()
         self._transaction = self._connection._connection.transaction()
+        logger.warning(self._connection)
 
         async with self._connection._transaction_lock:
             is_root = not self._connection._transaction_stack
--- a/databases/core.py
+++ b/databases/core.py
@@ -68,7 +68,9 @@ class Database:
         self._backend = backend_cls(self.url, **self.options)
 
         # Connections are stored as task-local state.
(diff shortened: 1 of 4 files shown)


ToGoBananas:
looks good, no?

erakli:
@tomchristie take a look at this PR please

databases/core.py
```diff
@@ -357,6 +372,7 @@ async def wrapper(*args: typing.Any, **kwargs: typing.Any) -> typing.Any:
     async def start(self) -> "Transaction":
         self._connection = self._connection_callable()
         self._transaction = self._connection._connection.transaction()
+        logger.warning(self._connection)
```
erakli, line 375: do we really need this warning?

databases/core.py
```diff
@@ -173,21 +175,34 @@ async def iterate(
...
 
     def transaction(
         self, *, force_rollback: bool = False, **kwargs: typing.Any
     ) -> "Transaction":
-        return Transaction(self.connection, force_rollback=force_rollback, **kwargs)
+        try: 
+            connection = self._connection_context.get()
+            if not connection._transaction_stack:
+                newcontext = contextvars.copy_context()
+                get_conn  = lambda: newcontext.run(self._new_connection)
```
erakli, line 199:
```suggestion
                get_conn = lambda: newcontext.run(self._new_connection)
```

databases/core.py
```diff
@@ -173,21 +175,34 @@ async def iterate(
...
-            return connection
+            return self._new_connection()
 
     def transaction(
         self, *, force_rollback: bool = False, **kwargs: typing.Any
     ) -> "Transaction":
-        return Transaction(self.connection, force_rollback=force_rollback, **kwargs)
+        try: 
+            connection = self._connection_context.get()
+            if not connection._transaction_stack:
```
erakli, line 197: and for readability 

```suggestion
            is_root = not connection._transaction_stack
            if is_root:
```

goteguru: yes, it may help.

databases/core.py
```diff
@@ -15,8 +15,10 @@
 
 if sys.version_info >= (3, 7):  # pragma: no cover
     from contextvars import ContextVar
+    import contextvars as contextvars
```
aminalaee, line 18: Can L17, L18 imports be merged?

aminalaee:
@goteguru Please run linting/formatting scripts in the `scripts` folder.

goteguru:
> > @tomchristie take a look at this PR please
> 
> I'm happy to help out with expanding the maintainers team as needed, but I don't have any time to invest in `databases` myself.

Thx for the honest reply. In the meantime I've replaced `databases` with direct `asyncpg` calls in my project (I don't like if local changes of dependencies unmerged to upstream), but I believe my future (concurrent)  project can profit from this fix. 

Maybe it would be nice to list the active project maintainers (w/ commit right) in the project page.

databases/core.py
goteguru: Is there a format to do this in a single line? :-o
Probably "explicit is better than implicit" principle should be applied here and ContextVar should be contextvars.ContextVar to emphasize the dependency.  Of course we could also import `copy_context` since this is the only function we need, but (i think) using foreign namespace directly is not very fortunate.

aminalaee: @goteguru 
Sorry if my comment was confusing :D
What I mean is that can we do `import contextvars` and get `contextvars.ContextVars` and `contextvars.copy_context`?

goteguru:
@aminalaee 
> @goteguru Please run linting/formatting scripts in the `scripts` folder.
I already did manually (all tests are passed), but I run them again for isort. 
BTW: some windows commit corrupted the line endings, therefore the scripts will fail on linux using dash for sh (eg. all ubuntu). 

Please consider recommended best practices while working from a windows machine:
[git:formatting and whitespace](http://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_formatting_and_whitespace)

databases/core.py
aminalaee: @goteguru what do you think about my comment?
If you don't have the time I can do that change, but your input would be appreciated. I look forward to get this through.

tests/test_databases.py
```diff
@@ -800,7 +800,7 @@ async def test_queries_with_expose_backend_connection(database_url):
     """
     async with Database(database_url) as database:
         async with database.connection() as connection:
-            async with database.transaction(force_rollback=True):
+            async with connection.transaction(force_rollback=True):
```
aminalaee, line 803: The docs for transactions need to be updated too.

aminalaee (changes requested):

aminalaee (approved):
Thank you @goteguru

tests/test_databases.py
goteguru: Sure. Thank you for the fix! (I'm glad it got merged finally).

PR #371, opened 2021-08-18
Use `executemany` from backends

### LINKED ISSUE #284: Inserting to database using execute_many much slower than using fetch_all
Hi,

I noticed that inserting multiple rows with execute_many is much slower than with fetch_all
```
"inserting with execute_many": "Time (hh:mm:ss.ms) 0:00:02.538554",
"inserting with fetch_all": "Time (hh:mm:ss.ms) 0:00:00.249998",
```

aminalaee:
Closes #284 

Some of the backends support `executemany` and instead of running multiple single queries we can use `executemany`.
asyncpg, aiosqlite and aiomysql support `executemany` but `aiopg` doesn't support it.

### DIFF
--- a/databases/backends/mysql.py
+++ b/databases/backends/mysql.py
@@ -144,10 +144,9 @@ class MySQLConnection(ConnectionBackend):
     async def execute_many(self, queries: typing.List[ClauseElement]) -> None:
         assert self._connection is not None, "Connection is not acquired"
         cursor = await self._connection.cursor()
+        query, values = self._compile_many(queries)
         try:
-            for single_query in queries:
-                single_query, args, context = self._compile(single_query)
-                await cursor.execute(single_query, args)
+            await cursor.executemany(query, values)
         finally:
             await cursor.close()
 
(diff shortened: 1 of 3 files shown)


databases/backends/mysql.py
```diff
@@ -213,6 +212,16 @@ def _compile(
         logger.debug("Query: %s Args: %s", query_message, repr(args), extra=LOG_EXTRA)
         return compiled.string, args, CompilationContext(execution_context)
 
+    def _compile_many(
+        self, queries: typing.List[ClauseElement]
+    ) -> typing.Tuple[str, typing.Sequence[dict]]:
```
PrettyWood, line 217:
```suggestion
    ) -> typing.Tuple[str, typing.List[dict]]:
```
IMO better be explicit in return value and as laxist as possible in args (Sequence, Mapping, ...)

PrettyWood (commented):
I would have expected a change in
https://github.com/encode/databases/blob/master/databases/core.py#L281-L286
instead of creating multiple queries to take again the first one afterwards
Do I miss something? 🤔

aminalaee:
@PrettyWood Yeah I actually thought that would be more efficient too. If I understand your comment correctly, you want to avoid building all the queries in `core` and you suggest we build only the first query and pass the values to the backends?

I tried that before and there's a limitation.
In each database backend we're doing a `_compile` which builds the query and parameters from a `ClauseElement` instance.
https://github.com/encode/databases/blob/master/databases/backends/sqlite.py#L157

If we build only the first query in `core.py`, and pass the values around to the database backends, we will be calling `executemany` method with a `compiled` query string which is correct but the paramters are not coming from a compiled query meaning they're not coming from the output of `_compile` method.

I'm really interested to improve this and get it through. So feel free to dig deeper and let me know if we can improve it.

aminalaee:
Closing it as stale.

PR #378, opened 2021-08-31
support `render_postcompile`

### LINKED ISSUE #377: SQLAChemy 1.4 POSTCOMPILE error
I try new encode/databases and  SQLAChemy 1.4 everything  work fine, except

```
sql = select([TABLE]).where(TABLE.c.id.in_([1,3,4]))
print(databases.fetch_all(sql))
```

it produced  error

```
psycopg2.errors.SyntaxError: syntax error at or near "["
LINE 3: WHERE TABLE.id IN ([POSTCOMPILE_1])
```

But I can use this query with pure SQLAChemy 1.4 core

```
sql = select([TABLE]).where(TABLE.c.id.in_([1,3,4]))
async with engine.connect() as conn:
      res = await conn.execute(sql)
print(res.fetchall())
```

aminalaee:
Fixes #377

The behavior of `IN` and `NOT IN` Bind parameters have changed to `POSTCOMPILE` which will expand parameters on the fly.

SQLAlchemy changelog [here](https://docs.sqlalchemy.org/en/14/changelog/migration_14.html#all-in-expressions-render-parameters-for-each-value-in-the-list-on-the-fly-e-g-expanding-parameters).

```python
stmt = select(A.id, A.data).where(A.id.in_([1, 2, 3]))

>>> print(stmt)
SELECT a.id, a.data
FROM a
WHERE a.id IN ([POSTCOMPILE_id_1])
```

We should either use `literal_binds` with compile:
```python
>>> print(stmt.compile(compile_kwargs={"literal_binds": True}))
SELECT a.id, a.data
FROM a
WHERE a.id IN (1, 2, 3)
```

Or use new `render_postcompile` helper:
```python
>>> print(stmt.compile(compile_kwargs={"render_postcompile": True}))
SELECT a.id, a.data
FROM a
WHERE a.id IN (:id_1_1, :id_1_2, :id_1_3)
```

### DIFF
--- a/databases/backends/aiopg.py
+++ b/databases/backends/aiopg.py
@@ -199,7 +199,9 @@ class AiopgConnection(ConnectionBackend):
     def _compile(
         self, query: ClauseElement
     ) -> typing.Tuple[str, dict, CompilationContext]:
-        compiled = query.compile(dialect=self._dialect)
+        compiled = query.compile(
+            dialect=self._dialect, compile_kwargs={"render_postcompile": True}
+        )
 
         execution_context = self._dialect.execution_ctx_cls()
         execution_context.dialect = self._dialect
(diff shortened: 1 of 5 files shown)


PrettyWood (approved):
Thanks for that!
I guess a 0.5.1 should be done

Issue #359, opened 2021-07-05
Connection and disconnection idempotence


sondrelg:
Hi!

First of all thanks for creating and maintaining this! 👏   

### Question

The [connect](https://github.com/encode/databases/blob/master/databases/core.py#L82) and [disconnect](https://github.com/encode/databases/blob/master/databases/core.py#L105) core methods have an assertion each preventing repeat calls:

```python
assert not self.is_connected, "Already connected."
```

Would it make sense to replace these with an early return, to allow several repeat calls to the methods? Like this?

```python
if self.is_connected:
    return
```

### Motivation

Since I couldn't find any other issues or discussions for this I'm concerned I might be misusing the library in some way, but my root issue is basically this: 

I have an app with a few `databases` async context managers in different parts of my application - when testing this logic I sometimes want to create DB-fixtures which also requires me to open a connection. As soon as these mix, things break, which makes things complicated seems unnecessary (but I'm sure it might not be 🙂 )

-----------------

tl;dr: Would it be harmful to remove the asserts? It would make my life *a little* easier at least 😄

Id be happy to submit a PR if there aren't any reason why this shouldn't be allowed 👍

aminalaee:
@sondrelg I see the point, but if you're still interested, can you provide some examples with this issue?
I don't see any harm in creating a PR.

sondrelg:
In our test-setup we do something like this to create database resources (Starlette project), with pytest fixtures:

```python
# tests/conftest.py
@pytest.fixture
async def _db(_db_schema):
    if db.database.is_connected:
        await db.database.disconnect()

    if not db.database.is_connected:
        await db.database.connect()

    yield

    if db.database.is_connected:
        await db.database.disconnect()

@pytest.fixture
async def create_resource(_db):
    return await Resource.create(**{dict values})
```

And we also have this in our worker's main function:

```python
# worker.py // main function
    ...
    if not db.database.is_connected:
        await db.database.connect()
    
    try:
        # fire off a while loop
    except ...:
        ...
    finally:
        if db.database.is_connected:
            await db.database.disconnect()
```

_Without_ the library asserts I believe we could do this instead: 

```python
# tests/conftest.py
@pytest.fixture
async def _db(_db_schema):
    async with database.connection() as connection:
        yield

@pytest.fixture
async def create_resource(_db):
    return await Resource.create(**{dict values})
```
```python
# worker.py // main function    
    ...
    async with database.connection() as connection:
    
        try:
            # fire off a while loop
[...]

PR #379, opened 2021-09-01
Allow repeated calls to connect and disconnect

sondrelg:
Replaces assertions with early returns as discussed in https://github.com/encode/databases/issues/359.

### DIFF
--- a/tests/test_databases.py
+++ b/tests/test_databases.py
@@ -737,6 +737,13 @@ async def test_connect_and_disconnect(database_url):
     await database.disconnect()
     assert not database.is_connected
 
+    # Test duplicate connect and disconnect
+    await database.connect()
+    await database.connect()
+    assert database.is_connected
+    await database.disconnect()
+    await database.disconnect()
+    assert not database.is_connected
 
 @pytest.mark.parametrize("database_url", DATABASE_URLS)
 @async_adapter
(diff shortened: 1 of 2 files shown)


sondrelg:
Looks like the test suite fails from a drop in coverage. I'll happily add tests for this, but do you think this is otherwise mergable seeing that no tests actually failed here @aminalaee? 🙂

aminalaee:
@sondrelg Thanks you for this. Yes, it looks good to me , and I don't think it would break anything.
I'll keep it open for a day or two to see if anyone has any objections though.

aminalaee:
Closes #359

sondrelg:
Ok cool, I'll add a test or two!

tests/test_databases.py
```diff
@@ -737,6 +737,13 @@ async def test_connect_and_disconnect(database_url):
     await database.disconnect()
     assert not database.is_connected
 
+    # Test duplicate connect and disconnect
+    await database.connect()
+    await database.connect()
+    assert database.is_connected
+    await database.disconnect()
+    await database.disconnect()
+    assert not database.is_connected
```
aminalaee, line 747:
```suggestion


```

aminalaee (changes requested):

aminalaee (approved):
@sondrelg  Thanks for this.

PR #382, opened 2021-09-06
mysql asyncmy driver

### LINKED ISSUE #244: Migrate away from aiomysql
`aiomysql` is outdated. The last release is almost 2 years old and there is no support for newer PyMySQL release like 0.10.1 which is the latest one.

There is not ETA for a new release (https://github.com/aio-libs/aiomysql/issues/466) or support for PyMySQL > 0.10.

`aiomysql` is a dependency of `databases` and `databases` is dependency of various web framework for DB support. In the current situation all those web frameworks don't have MySQL support on Fedora (Fedora ships  PyMySQL 0.10) and likely other distributions when installed with the distribution-provided packages.

It would be highly appreciated if you consider to migrate to MySQl binding which are actively maintained.

Thanks

aminalaee:
Fixes #244 .
Adding [asyncmy](https://github.com/long2ice/asyncmy) as another mysql driver which is more up to date and faster in some areas.

- [x]  Adding asyncmy
- [x] Fixing all tests (raw queries are failing)
- [x] Update the docs

### DIFF
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -14,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.6", "3.7", "3.8"]
+        python-version: ["3.7", "3.8", "3.9"]
 
     services:
       mysql:
--- a/.github/workflows/test-suite.yml
+++ b/.github/workflows/test-suite.yml
@@ -14,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.7", "3.8", "3.9"]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10"]
 
     services:
       mysql:
@@ -29,7 +29,7 @@ jobs:
         options: --health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=3
 
       postgres:
-        image: postgres:10.8
+        image: postgres:14
         env:
           POSTGRES_USER: username
           POSTGRES_PASSWORD: password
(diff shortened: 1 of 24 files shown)


aminalaee:
@dkharyna I'm waiting for feedback.

gordthompson (approved):
This seems to work fine for me in my little FastAPI test rig at

https://github.com/gordthompson/fastapi-crud-demo

I would prefer that `mysql://…` defaulted to asyncmy instead of aiomysql, but I understand that you might not want to do that right away for fear of breaking stuff. Maybe eventually? (It really does look like the whole "aio*" family of drivers has been orphaned.)

aminalaee:
@gordthompson Thank you for the review.

TBH, I think it would be better to just remove the default database drivers. They can be confusing, we can stick to explicit drivers.

aminalaee:
@tomchristie Friendly reminder :)

tomchristie (approved):
I'm just going to leave these ones up to your call @aminalaee.

Generally worth trying to keep different aspects into different PRs (Eg. I notice there's some changes to the development scripts here, which really could be kept separate to the `asyncmy` driver addition.) But I'm not bothered enough about that to block this from being able to go in, if you think it's ready.

aminalaee:
> I'm just going to leave these ones up to your call @aminalaee.
> 
> Generally worth trying to keep different aspects into different PRs (Eg. I notice there's some changes to the development scripts here, which really could be kept separate to the `asyncmy` driver addition.) But I'm not bothered enough about that to block this from being able to go in, if you think it's ready.

Thanks for the review. You're right I could leave a little bit of that out of this MR, but as you said it wasn't that much.
Basically I had to edit the actions to allow python coverage for python 3.10 and thought now that I'm here, I'd rather fix this.

Will keep that in mind.

tomchristie:
Yup. To be fair - I do *sometimes* do the same myself. 🤭

PR #385, opened 2021-09-09
Failed connection counter reset

taybin:
This adds a unittest for #283.  It does require the library `asyncmock` as AsyncMock isn't available in python 3.6 or 3.7 yet.  If testing on 3.6 and 3.6 is not a requirement, we can use unittest.mocks.AsyncMock since it's included in python 3.8.

@aminalaee

### DIFF
--- a/databases/core.py
+++ b/databases/core.py
@@ -237,8 +237,12 @@ class Connection:
     async def __aenter__(self) -> "Connection":
         async with self._connection_lock:
             self._connection_counter += 1
-            if self._connection_counter == 1:
-                await self._connection.acquire()
+            try:
+                if self._connection_counter == 1:
+                    await self._connection.acquire()
+            except Exception as e:
+                self._connection_counter -= 1
+                raise e
         return self
 
     async def __aexit__(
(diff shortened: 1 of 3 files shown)


taybin:
I tried the test with and without the code fix and the test fails as expected without the _connection_counter rollback.

aminalaee:
@taybin Thanks for this. Can we remove `asyncmock` requirement with this?

```python
class AsyncMock(MagicMock):
    async def __call__(self, *args, **kwargs):
        return super(AsyncMock, self).__call__(*args, **kwargs)
```

[credit](https://stackoverflow.com/questions/32480108/mocking-async-call-in-python-3-5)

aminalaee:
@taybin I think the `Exception("boo!")` can be simplified to `Exception()` too. We're basically comparing generated exception with itself.

taybin:
@aminalaee I wanted to be sure it was that specific Exception that was being caught.  `pytest.raises(Exception)` will pass on any error and  I wanted to avoid that.  How's this change?  (Removed the asyncmock dep too)

aminalaee:
@taybin I was thinking raising an exception like `TimeoutError` and comparing that instead of `Exception`. But I think what we already have is great. Thanks.

aminalaee (approved):

PR #390, opened 2021-09-12
fix documentation about low level transaction API usage

### LINKED ISSUE #389: fix documentation about low level transaction API usage
According to the [encode/Databases documentation](https://www.encode.io/databases/connections_and_transactions/#transactions) about Transactions, a common pattern for lower-level transaction API is:

```python
transaction = await database.transaction()
try:
    await transaction.start()
    ...
except:
    await transaction.rollback()
else:
    await transaction.commit()
```

However, calling `await transaction.start()` is a duplication and leads to error because it is implicitly invoked by `await database.transaction()`, as can be seen [from this snippet of code](https://github.com/encode/databases/blob/2150a2255356f0c1a529fc5363af0c4cbe7b92f3/databases/core.py#L365:L369:).

So, the documentation should skip calling `start`:

```python
transaction = await database.transaction()
try:
    ...
except:
    await transaction.rollback()
else:
    await transaction.commit()
```

[In the chat room](https://gitter.im/encode/community), @aminalaee pointed out it would be possible to use `start` too if the `transaction` instantiation wasn't `awaited`:

```python
transaction = database.transaction()
await transaction.start()
try:
    ...
except:
    await transaction.rollback()
else:
    await transaction.commit()
```

Although both forms are correct, keeping them both in the documentation will be confusing. I recommend that we keep the first one because it is simpler.

andredias:
fixes #389

### DIFF
--- a/docs/connections_and_transactions.md
+++ b/docs/connections_and_transactions.md
@@ -80,7 +80,6 @@ For a lower-level transaction API:
 ```python
 transaction = await database.transaction()
 try:
-    await transaction.start()
     ...
 except:
     await transaction.rollback()


aminalaee (approved):
Thanks for this.

PR #408, opened 2021-10-15
change Record type to Sequence

### LINKED ISSUE #407: Postgres backend Record is a Mapping but some Mapping methods are deprecated
Since https://github.com/encode/databases/pull/299 upgraded to sqlalchemy 1.4, the postgres backend's Record object now mimics the behavior of sqlalchemy's Row which is meant to behave similarly to a NamedTuple (and inherits from collections.abc.Sequence) https://docs.sqlalchemy.org/en/14/changelog/migration_14.html#change-4710-core

Meanwhile, postgres backend's Record object inherits from collections.abc.Mapping and is therefore required to fulfill the Mapping interface, which includes keys() and values() which are now deprecated.

Sqlalchemy provides a `mapping()` method on Result which will cause it to return RowMapping objects rather than Row objects, and those look like Mappings.

I encountered this issue working with fastapi and pydantic. Returning Records as pydantic models worked in the past, but now produces a deprecation warning (and i guess will eventually stop working) since pydantic's builtin validator treats the Record as a Mapping and attempts to call `dict(record)`

aminalaee:
Fixes #407 .

https://docs.sqlalchemy.org/en/14/changelog/migration_14.html#change-4710-core

Since SQLAlchemy 1.4 the Row object deprecates `.keys()` `.values()` and `.items()` and they're exposed through `Row._mapping`.

There's no point in keeping the `Mapping` and instead change that to `Sequence` to be the same as `Row`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,6 @@ setup(
     author_email="tom@tomchristie.com",
     packages=get_packages("databases"),
     package_data={"databases": ["py.typed"]},
-    data_files=[("", ["LICENSE.md"])],
     install_requires=['sqlalchemy>=1.4,<1.5', 'aiocontextvars;python_version<"3.7"'],
     extras_require={
         "postgresql": ["asyncpg"],


peku33:
Wasn't that used for raw sql queries (not relaying) on sqlalchemy?
For mixed projects (some queries via sqlalchemy, some via sql) this makes a huge breaking change.

aminalaee:
It's already mentioned here: #445

Issue #413, opened 2021-10-27
feature request: iterate over more than 1 record


databasedav:
something like this
```python
async for rows in database.iterate(query, values, n=10):
    rows
```
where `rows` would have 10 rows until there is <10 to be returned and it would simply return the rest

sobolevn:
I think that a proper way to do that is to use something like https://aioitertools.omnilib.dev/en/stable/api.html#aioitertools.itertools.islice

PR #414, opened 2021-10-27
allow iterating over custom number of records with asyncpg

databasedav:
addresses https://github.com/encode/databases/issues/413, but only for postgres with asyncpg, passing `n` in the kwargs to `.iterate` will error for the other drivers, i can add them eventually if that's a requirement to get this merged

- [x] asyncpg
- [ ] tests
- [ ] other drivers
- [ ] documentation

### DIFF
--- a/databases/backends/aiopg.py
+++ b/databases/backends/aiopg.py
@@ -190,7 +190,9 @@ class AiopgConnection(ConnectionBackend):
                 metadata._keymap,
                 Row._default_key_style
             )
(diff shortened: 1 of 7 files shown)


aminalaee:
Any update with this?

databasedav:
i was having issue with running these concurrently, i think it was related to the transaction issues mentioned in the repo, i am now just using sqlalchemy since it has asyncpg support

PR #416, opened 2021-10-27
"Connection is already acquired" error

bokolob:


bokolob:
I've written a new version of the tests, and they reproduce the error. But, the suit hangs up after mysql/postgres tests finish. No idea why.

databases/backends/mysql.py
```diff
@@ -97,9 +97,12 @@ async def acquire(self) -> None:
     async def release(self) -> None:
         assert self._connection is not None, "Connection is not acquired"
         assert self._database._pool is not None, "DatabaseBackend is not running"
-        await self._database._pool.release(self._connection)
+        await self._internal_release(self._connection)
```
aminalaee, line 100: Sorry I'm confused, how is this different from the original one?

databases/backends/postgres.py
```diff
@@ -263,7 +267,7 @@ def _compile(self, query: ClauseElement) -> typing.Tuple[str, list, tuple]:
 
     @staticmethod
     def _create_column_maps(
-        result_columns: tuple,
+            result_columns: tuple,
```
aminalaee, line 270: I think it's mixing tab and spaces in multiple files. You can use `./scripts/lint` to fix that.

databases/backends/mysql.py
bokolob: It seems, I forgot to make changes for mysql during debugging

aminalaee:
Any update with this?

aminalaee:
I think this is out of date now. Closing it.

bokolob:
Hello, no it’s still in progress. I haven’t had enough time for this. So you can help with tests :)

Beaend:
i got this error then write " instead of '
`SELECT {need} FROM {table_name} WHERE {column_name} = "{column_text}"`
first i got error `UndefinedColumnError`, then this.
(i'm use pool)

PR #436, opened 2021-12-10
Change extra installations to specific drivers

### LINKED ISSUE #421: Consider deprecating (or at least downplaying) the notion of a "default driver"
Further to comment from @aminalaee on #382 :

https://github.com/encode/databases/pull/382#issuecomment-955175931

Consider moving away from the notion of a "default driver" and encourage users to specify the driver that they actually want to use. In #382 it is asyncmy vs. aiomysql, but with the release of [psycopg 3.x](https://pypi.org/project/psycopg/) is it also an issue for PostgreSQL since the new driver also supports async (so asyncpg vs. psycopg).

I would be happy to work on a PR if that would further the cause.

### LINKED ISSUE #429: broken installation instructions for non-default db drivers
Via readme.md:
```
You can also use other database drivers supported by databases:

$ pip install databases[postgresql+aiopg]
$ pip install databases[mysql+asyncmy]
```

Is there any python/pip combination for which this works? I swear I've seen the same plus notation before somewhere but I'm pretty sure it doesn't work:

```
docker run --rm python:3.9 bash -exc "pip install -U pip==21.3.1; pip install 'databases[postgresql+aiopg]'"

+ pip install -U pip==21.3.1
Collecting pip==21.3.1
  Downloading pip-21.3.1-py3-none-any.whl (1.7 MB)
Installing collected packages: pip
  Attempting uninstall: pip
    Found existing installation: pip 21.2.4
    Uninstalling pip-21.2.4:
      Successfully uninstalled pip-21.2.4
Successfully installed pip-21.3.1
WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv
+ pip install 'databases[postgresql+aiopg]'
ERROR: Exception:
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/pip/_vendor/packaging/requirements.py", line 102, in __init__
    req = REQUIREMENT.parseString(requirement_string)
  File "/usr/local/lib/python3.9/site-packages/pip/_vendor/pyparsing.py", line 1955, in parseString
    raise exc
  File "/usr/local/lib/python3.9/site-packages/pip/_vendor/pyparsing.py", line 3814, in parseImpl
    raise ParseException(instring, loc, self.errmsg, self)
pip._vendor.pyparsing.ParseException: Expected stringEnd, found '['  (at char 11), (line:1, col:12)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/pip/_internal/cli/base_command.py", line 164, in exc_logging_wrapper
    status = run_func(*args)
  File "/usr/local/lib/python3.9/site-packages/pip/_internal/cli/req_command.py", line 205, in wrapper
    return func(self, options, args)
  File "/usr/local/lib/python3.9/site-packages/pip/_internal/commands/install.py", line 305, in run
    reqs = self.get_requirements(args, options, finder, session)
  File "/usr/local/lib/python3.9/site-packages/pip/_internal/cli/req_command.py", line 380, in get_requirements
    req_to_add = install_req_from_line(
  File "/usr/local/lib/python3.9/site-packages/pip/_internal/req/constructors.py", line 366, in install_req_from_line
    parts = parse_req_from_line(name, line_source)
  File "/usr/local/lib/python3.9/site-packages/pip/_internal/req/constructors.py", line 306, in parse_req_from_line
    extras = convert_extras(extras_as_string)
  File "/usr/local/lib/python3.9/site-packages/pip/_internal/req/constructors.py", line 58, in convert_extras
    return get_requirement("placeholder" + extras.lower()).extras
  File "/usr/local/lib/python3.9/site-packages/pip/_internal/utils/packaging.py", line 84, in get_requirement
    return Requirement(req_string)
  File "/usr/local/lib/python3.9/site-packages/pip/_vendor/packaging/requirements.py", line 104, in __init__
    raise InvalidRequirement(
pip._vendor.packaging.requirements.InvalidRequirement: Parse error at "'[postgre'": Expected stringEnd
```

aminalaee:
Closes #429 and closes #421.

Changes extra installations from:

```shell
pip install databases[postgresql+asyncpg]
```

To:

```shell
pip install databases[asyncpg]
```

This will be a step towards downplaying the default drivers.
Even though it keeps default drivers to work like `pip install databases[postgresql]` but will remove them from the docs to point to specific drivers.

And also worth mentioning that this will be the API, still mentioned the default in the docs:

```python
database = Database('sqlite:///example.db')
```

This is equivalent to:

```python
database = Database('sqlite+aiosqlite:///example.db')
```

I think we should also downplay the default dialect and specify the driver in the docs.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -27,6 +27,14 @@ Databases is suitable for integrating against any async Web framework, such as [
 $ pip install databases
 ```
 
+Database drivers supported are:
+
+* [asyncpg][asyncpg]
+* [aiopg][aiopg]
+* [aiomysql][aiomysql]
+* [asyncmy][asyncmy]
+* [aiosqlite][aiosqlite]
+
 You can install the required database drivers with:
 
 ```shell
@@ -59,7 +67,7 @@ expressions directly from the console.
 ```python
 # Create a database instance, and connect to it.
 from databases import Database
-database = Database('sqlite:///example.db')
+database = Database('sqlite+aiosqlite:///example.db')
 await database.connect()
 
 # Create a table.
@@ -93,6 +101,11 @@ for examples of how to start using databases together with SQLAlchemy core expre
 [alembic]: https://alembic.sqlalchemy.org/en/latest/
 [psycopg2]: https://www.psycopg.org/
 [pymysql]: https://github.com/PyMySQL/PyMySQL
+[asyncpg]: https://github.com/MagicStack/asyncpg
+[aiopg]: https://github.com/aio-libs/aiopg
+[aiomysql]: https://github.com/aio-libs/aiomysql
+[asyncmy]: https://github.com/long2ice/asyncmy
+[aiosqlite]: https://github.com/omnilib/aiosqlite
 
 [starlette]: https://github.com/encode/starlette
 [sanic]: https://github.com/huge-success/sanic
--- a/docs/index.md
+++ b/docs/index.md
@@ -93,9 +93,6 @@ for examples of how to start using databases together with SQLAlchemy core expre
 [alembic]: https://alembic.sqlalchemy.org/en/latest/
 [psycopg2]: https://www.psycopg.org/
 [pymysql]: https://github.com/PyMySQL/PyMySQL
-[asyncpg]: https://github.com/MagicStack/asyncpg
-[aiomysql]: https://github.com/aio-libs/aiomysql
-[aiosqlite]: https://github.com/jreese/aiosqlite
 
 [starlette]: https://github.com/encode/starlette
 [sanic]: https://github.com/huge-success/sanic
--- a/docs/index.md
+++ b/docs/index.md
@@ -30,18 +30,11 @@ $ pip install databases
 You can install the required database drivers with:
 
 ```shell
-$ pip install databases[postgresql]
-$ pip install databases[mysql]
-$ pip install databases[sqlite]
-```
-
-Default driver support is provided using one of [asyncpg][asyncpg], [aiomysql][aiomysql], or [aiosqlite][aiosqlite].
-
-You can also use other database drivers supported by `databases`:
-
-```shel
-$ pip install databases[postgresql+aiopg]
-$ pip install databases[mysql+asyncmy]
+$ pip install databases[asyncpg]
+$ pip install databases[aiopg]
+$ pip install databases[aiomysql]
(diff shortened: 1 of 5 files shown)


aminalaee:
@collerek Do you think you can review this?

docs/index.md
```diff
@@ -100,9 +93,6 @@ for examples of how to start using databases together with SQLAlchemy core expre
 [alembic]: https://alembic.sqlalchemy.org/en/latest/
 [psycopg2]: https://www.psycopg.org/
 [pymysql]: https://github.com/PyMySQL/PyMySQL
-[asyncpg]: https://github.com/MagicStack/asyncpg
```
collerek, line 103: I would leave the links and add new links to rest of the drivers. It should allow people to choose a driver they want, especially for backends that support more than one (MySQL and postgres)

setup.py
```diff
@@ -50,10 +50,13 @@ def get_packages(package):
...
     extras_require={
         "postgresql": ["asyncpg"],
+        "asyncpg": ["asyncpg"],
+        "aiopg": ["aiopg"],
         "mysql": ["aiomysql"],
-        "mysql+asyncmy": ["asyncmy"],
+        "aiomysql": ["aiomysql"],
+        "asyncmy": ["asyncmy"],
         "sqlite": ["aiosqlite"],
-        "postgresql+aiopg": ["aiopg"],
```
collerek, line 56: This will break pinned dependencies so I would change version of the lib to 0.6.0 in this commit.

That way, when the new release will be made you won't forget to bump the minor number. 

The package is published to pypi only with a new release anyway so it should be fine to bump the version upfront.

aminalaee:
@collerek thanks for the review. that makes sense.
I'll also update documentation for the database URI to downplay the idea of default drivers:

So instead of this:

```python
database = Database('sqlite:///example.db')
```

We would use this in the docs:

```python
database = Database('sqlite+aiosqlite:///example.db')
```

Even though both will work for now, but we shouldn't;t encourage the default drivers both in installation and usage.

setup.py
aminalaee: I created a separate PR for this as we'll need to update changelog too.

collerek (approved):

PR #449, opened 2022-01-20
Temporary PR for test suit

collerek:
Tests for 0.5.5. release.

@aminalaee Here you go

BTW. I want to make a PR with some docker config so you can test all backends at localhost without uploading.


aminalaee:
Looks good 👍

collerek:
Released

Issue #196, opened 2020-04-26
Support SQLite URI mode and arguments


brendanlong:
## Background

In sqlite3, if you [set `uri=True`](https://docs.python.org/3/library/sqlite3.html#sqlite3.connect), you can pass query param arguments to the underlying database, including [options like `cache=shared` and `mode=memory`](https://www.sqlite.org/inmemorydb.html#sharedmemdb), which are useful for unit testing (you can create an in-memory DB which stays alive between multiple connections).

For example, without the query params, `:memory:` databases are completely separate:

```
import sqlite3

db1 = sqlite3.connect(":memory:")
db1.execute("CREATE TABLE Example(id INT)")
db1.execute("SELECT * FROM Example")

db2 = sqlite3.connect(":memory:")
db2.execute("SELECT * FROM Example") # sqlite3.OperationalError: no such table: Example
```

But if we pass `cache=shared`, they're shared:

```
import sqlite3

db1 = sqlite3.connect("file::memory:?cache=shared", uri=True)
db1.execute("CREATE TABLE Example(id INT)")
db1.execute("SELECT * FROM Example")

db2 = sqlite3.connect("file::memory:?cache=shared", uri=True)
db2.execute("SELECT * FROM Example") # no error
```

`mode=memory` lets you create named memory DB's so you can share them by name instead of having a single shared namespace.

## The problem

The problem is that `databases` handles query params itself and doesn't pass them to the underlying connection:

```
import asyncio
import sqlite3

from databases import Database

async def example():
    _ = sqlite3.connect("file::memory:?cache=shared", uri=True) # Hold a reference open since SQLite deletes the shared DB when ths last reference closes
    db1 = Database("sqlite:///file::memory:?cache=shared", uri=True)
    await db1.execute("CREATE TABLE Example(id INT)")
    await db1.execute("SELECT * FROM Example")
    db2 = Database("sqlite:///file::memory:?cache=shared", uri=True)
    await db2.execute("SELECT * FROM Example") # sqlite3.OperationalError: no such table: Example

asyncio.run(example())
```

## Proposed solutions

I messed around with this locally and come up with this solution:

```
diff --git a/databases/backends/sqlite.py b/databases/backends/sqlite.py
index 6a733ac..7f46c76 100644
--- a/databases/backends/sqlite.py
+++ b/databases/backends/sqlite.py
@@ -1,5 +1,6 @@
 import logging
 import typing
+from urllib.parse import urlencode
 import uuid
 
 import aiosqlite
[...]

Issue #75, opened 2019-04-02
in-memory sqlite odd behavior


Lawouach:
Hi,

Thanks for the awesome package (happy to use it with starlette). For rapid dev purpose, I'm relying on the SQLite backend and I see an odd behavior.

When I create tables with an in-memory db, those tables can never be found in sqlite_master afterwards. But when I rely on a file database, then that works fine.

Is there any gotcha for in-memory sqllite I should be aware of?

Cheers,

tomchristie:
Well, in memory will only be persisted for the duration of the python session.

You might also want to double check against what behaviour sqlite3 has. Does “creating the tables” for an in memory DB actually do *anything*?

tomchristie:
I’m going to close this off - if you thing there’s a active issue, and (ideally) can build a test case for it then we could reconsider.

Lawouach:
I meant within a python session.
Inside a given Python session, when using in-memory, the tables don't show up once created.

> Does “creating the tables” for an in memory DB actually do anything?

I'm not sure if you are being sarcastic or actually asking a question.

Fair enough. Leave it close.

tomchristie:
I’m not. I’d suggest start by comparing sqlite3’s behaviour vs ours. If you’ve double checked that it’s definitely different, and can give a minimal replicable example then i’ll happily reopen and put my time into it. I’d just like to have a clear confirmation that there’s an issue this side, before commiting time in.

Lawouach:
Here is an example of what I mean:

```import asyncio

from databases import Database
import sqlalchemy


metadata = sqlalchemy.MetaData()

notes = sqlalchemy.Table(
    "notes",
    metadata,
    sqlalchemy.Column("id", sqlalchemy.String(length=100)),
    sqlalchemy.Column("text", sqlalchemy.String(length=100))
)

database = Database('sqlite:///:memory:')


async def run():
    await database.connect()

    await database.execute("""
    CREATE TABLE notes (
        id CHAR(32) NOT NULL, 
        user_id CHAR(100) NOT NULL
    );""")

    r = await database.fetch_all("SELECT name FROM sqlite_master WHERE type='table' ORDER BY name")

    # prints an empty list when in-memory, otherwise it shows the notes table
    print(r)

    await database.disconnect()


if __name__ == '__main__':
    loop = asyncio.get_event_loop()
    loop.run_until_complete(run())
    loop.close()
```

If you switch to a file db, then the table will be printed. In-memory will not have the table listed.

[...]

PR #451, opened 2022-01-24
Open up SQLite query param, cached database functionality, inspired by #196

nvjoshi2:
## Background
In sqlite3, if you set `uri=True` you can pass query param arguments to the underlying database, including:

- `mode = memory, ro, rw`
- `cache = shared`

Examples:
```python
# Open a database in read-only mode.
con = sqlite3.connect("file:template.db?mode=ro", uri=True)

# Don't implicitly create a new database file if it does not already exist.
# Will raise sqlite3.OperationalError if unable to open a database file.
con = sqlite3.connect("file:nosuchdb.db?mode=rw", uri=True)

# Create a cached in-memory database.
con = sqlite3.connect("file::memory:?cache=shared", uri=True)

# Create independent named in-memory database.
con1 = sqlite3.connect("file:mem1?mode=memory&cache=shared", uri=True)
con2 = sqlite3.connect("file:mem2?mode=memory&cache=shared", uri=True)
```
Options like `cache=shared` and `mode=memory` are particularly useful for unit testing (you can create an in-memory DB which stays alive between multiple connections).

## The Issue
The issue is twofold: 

1. `databases` strips out query params before passing the connection  string to underling connection
2. `databases` closing the db connection after each query means that there is no living reference to a connection of the cached in-memory database once the query finishes. This causes the cached in-memory database to be deleted after each query even if we could set `cache=shared`

This leads to the following behavior:

```python
import asyncio
import sqlite3

async def example():
    database = Database("sqlite:///file::memory:?cache=shared", uri=True)
    await database.connect()
    await database.execute("CREATE TABLE Example(id INT)")
    rows = await database.fetch_all("SELECT * FROM Example") # throws sqlite3.OperationalError: no such table: Example

asyncio.run(example())
```

## Proposed Solution
```diff
diff --git a/databases/backends/sqlite.py b/databases/backends/sqlite.py
index 46a3951..6335465 100644
--- a/databases/backends/sqlite.py
+++ b/databases/backends/sqlite.py
@@ -1,6 +1,8 @@
 import logging
+import sqlite3
 import typing
 import uuid
+from urllib.parse import urlencode
 
 import aiosqlite
 from sqlalchemy.dialects.sqlite import pysqlite
@@ -40,7 +42,9 @@ class SQLiteBackend(DatabaseBackend):
         # )
 
     async def disconnect(self) -> None:
-        pass
+        # if it extsis, remove reference to connection to cached in-memory database on disconnect
+        if self._pool._memref:
+            self._pool._memref = None
         # assert self._pool is not None, "DatabaseBackend is not running"
         # self._pool.close()
         # await self._pool.wait_closed()
@@ -52,12 +56,20 @@ class SQLiteBackend(DatabaseBackend):
 
 class SQLitePool:
     def __init__(self, url: DatabaseURL, **options: typing.Any) -> None:
-        self._url = url
+        self._database = url.database
+        self._memref = None
+        # add query params to database connection string
+        if url.options:
+            self._database += "?" + urlencode(url.options)
         self._options = options
 
+        if url.options and "cache" in url.options:
+            # reference to a connection to the cached in-memory database must be held to keep it from being deleted
+            self._memref = sqlite3.connect(self._database, **self._options)
+
     async def acquire(self) -> aiosqlite.Connection:
         connection = aiosqlite.connect(
-            database=self._url.database, isolation_level=None, **self._options
+            database=self._database, isolation_level=None, **self._options
         )
         await connection.__aenter__()
         return connection
```

This allows query params to be passed to the underlying connection and, when `cache=shared`, a reference to a connection to the cached in-memory database to be held (allowing it to persist) until `Database.disconnect` is called, leading to the following good behavior: 

```python
import asyncio
import sqlite3

async def example():
    database = Database("sqlite:///file::memory:?cache=shared", uri=True)
    await database.connect()
    await database.execute("CREATE TABLE Example(id INT)")
    rows = await database.fetch_all("SELECT * FROM Example") #no error
    datbase.disconnect()

    database = Database("sqlite:///file::memory:?cache=shared", uri=True)
    await database.connect()
    rows = await database.fetch_all("SELECT * FROM Example") # throws sqlite3.OperationalError: no such table: Example

asyncio.run(example())
```

Also fixes: #196, #75

### DIFF
--- a/databases/backends/sqlite.py
+++ b/databases/backends/sqlite.py
@@ -1,6 +1,8 @@
 import logging
+import sqlite3
 import typing
 import uuid
+from urllib.parse import urlencode
 
 import aiosqlite
 from sqlalchemy.dialects.sqlite import pysqlite
@@ -40,7 +42,9 @@ class SQLiteBackend(DatabaseBackend):
         # )
 
     async def disconnect(self) -> None:
-        pass
+        # remove reference to cached database connection on disconnect
+        if self._pool._memref:
+            self._pool._memref = None
         # assert self._pool is not None, "DatabaseBackend is not running"
         # self._pool.close()
         # await self._pool.wait_closed()
@@ -52,12 +56,19 @@ class SQLiteBackend(DatabaseBackend):
 
 class SQLitePool:
     def __init__(self, url: DatabaseURL, **options: typing.Any) -> None:
-        self._url = url
+        self._database = url.database
+        self._memref = None
+        if url.options:
+            self._database += "?" + urlencode(url.options)
         self._options = options
+
+        if url.options and "cache" in url.options:
+            # reference to cached memory database must be held to keep it from being deleted
+            self._memref = sqlite3.connect(self._database, **self._options)
 
     async def acquire(self) -> aiosqlite.Connection:
         connection = aiosqlite.connect(
-            database=self._url.database, isolation_level=None, **self._options
+            database=self._database, isolation_level=None, **self._options
         )
         await connection.__aenter__()
         return connection
--- a/databases/backends/sqlite.py
+++ b/databases/backends/sqlite.py
@@ -58,6 +58,7 @@ class SQLitePool:
     def __init__(self, url: DatabaseURL, **options: typing.Any) -> None:
         self._database = url.database
         self._memref = None
+        # add query params to database connection string
         if url.options:
             self._database += "?" + urlencode(url.options)
         self._options = options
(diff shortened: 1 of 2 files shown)


aweigold:
I tested this locally and this is really useful, especially when using databases and wanting to run unit tests in an in memory database with sqlite.

Are there any concerns with this PR?  Perhaps I could jump in and help address them.  It would be a shame to have to maintain a fork just for this feature.

farridav:
Any chance of a rebase @nvjoshi2 (presuming thats the reason this has not moved forwards)

zanieb:
@farridav I actually implemented this in #561 it should be available.

Issue #455, opened 2022-01-25
"Connection is not acquired"


dswillr:
I'm seeing `"Connection is not acquired"` errors being raised randomly in my multi-task asyncio application, even though `database.is_connected` is `True` on `databases==5.5.5` with `aiopg==1.3.3` and `python==3.10.2`.

```
  ... my code
    result = await database.fetch_one(query)
  File "/usr/local/lib/python3.10/site-packages/databases/core.py", line 155, in fetch_one
    return await connection.fetch_one(query, values)
  File "/usr/local/lib/python3.10/site-packages/databases/core.py", line 278, in fetch_one
    return await self._connection.fetch_one(built_query)
  File "/usr/local/lib/python3.10/site-packages/databases/backends/aiopg.py", line 137, in fetch_one
    assert self._connection is not None, "Connection is not acquired"
AssertionError: Connection is not acquired
```

Could this be an odd side effect of having connections shared across tasks ( https://github.com/encode/databases/issues/230 )?

dswillr:
Can confirm, sharing connections across multiple tasks is the issue. Replicated like so ( adjusted from #230 ):

```python
import asyncio
from databases import Database
database = Database("postgresql+aiopg://example:password@localhost/example")

async def t1():
    await database.connect()
    c1 = database.connection()

    async def inner():
        c2 = database.connection()
        value = await c2.fetch_val("select 1")  # raises "Connection is not acquired"
    await asyncio.gather(inner())


if __name__ == "__main__":
    loop = asyncio.get_event_loop()
    loop.run_until_complete(t1())
```

PR #457, opened 2022-02-01
Fix connections getting into a bad state when a task is cancelled

dswillr:
Resolves https://github.com/encode/databases/issues/455

### DIFF
--- a/tests/test_databases.py
+++ b/tests/test_databases.py
@@ -303,24 +303,26 @@ async def test_ddl_queries(database_url):
             await database.execute(query)
 
 
+class DBException(Exception):
+    pass
+
+
+@pytest.mark.parametrize("exception", [DBException, asyncio.CancelledError])
 @pytest.mark.parametrize("database_url", DATABASE_URLS)
 @mysql_versions
 @async_adapter
-async def test_queries_after_error(database_url):
+async def test_queries_after_error(database_url, exception):
     """
     Test that the basic `execute()` works after a previous error.
     """
-
-    class DBException(Exception):
-        pass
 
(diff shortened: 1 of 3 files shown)


tests/test_databases.py
```diff
@@ -303,24 +303,26 @@ async def test_ddl_queries(database_url):
             await database.execute(query)
 
 
+class DBException(Exception):
+    pass
```
aminalaee, line 307: Maybe we don't need this?

tests/test_databases.py
```diff
@@ -303,24 +303,26 @@ async def test_ddl_queries(database_url):
...
+
+
+@pytest.mark.parametrize("exception", [DBException, asyncio.CancelledError])
```
aminalaee, line 310: Could we do this?

```suggestion
@pytest.mark.parametrize("exception", [Exception, asyncio.CancelledError])
```

aminalaee (commented):
Thank you for the PR. Looks reasonable to me.

tests/test_databases.py
dswillr: I just kept what we had before, you're right we could remove that!

dswillr: I've applied your patch & removed the redundant class

aminalaee: Great, thanks.

aminalaee (approved):
🎉

kevburnsjr:
Any chance you might tag a new release so we can start using this fix?

aminalaee:
It will be in version 0.6.0.

PR #468, opened 2022-02-22
use execute_many from backends directly

Mng-dev-ai:


Mng-dev-ai:
@aminalaee Any updates?

databases/core.py
```diff
@@ -278,7 +278,7 @@ async def execute_many(
     ) -> None:
         queries = [self._build_query(query, values_set) for values_set in values]
         async with self._query_lock:
-            await self._connection.execute_many(queries)
+            await self._connection.execute_many(queries, values)
```
zanieb, line 281: Should we avoid building the queries here if you're going to compile them later?

databases/backends/postgres.py
```diff
@@ -216,14 +216,12 @@ async def execute(self, query: ClauseElement) -> typing.Any:
         query_str, args, result_columns = self._compile(query)
         return await self._connection.fetchval(query_str, *args)
 
-    async def execute_many(self, queries: typing.List[ClauseElement]) -> None:
+    async def execute_many(
+        self, queries: typing.List[ClauseElement], values: typing.List[dict]
+    ) -> None:
```
zanieb, line 221: I presume this it not a breaking change for users because this is an internal method and `values` is always provided by us in the public method?

databases/core.py
Mng-dev-ai: Hmm, it's been a long time since I wrote this PR, but I think we still need to convert the query to ClauseElement and bind params, but maybe we don't need list comprehension here, so maybe something like that is sufficient
```
async def execute_many(
        self, query: typing.Union[ClauseElement, str], values: list
    ) -> None:
        query = self._build_query(query,values[0])
        async with self._query_lock:
            await self._connection.execute_many(query, values)
```

databases/backends/postgres.py
Mng-dev-ai: I'm not sure I understand what you mean here, but I think users are only supposed to use execute_many from the core file, not from any of the backends

zanieb: Yep that's the question — I want to gauge if changing this function signature is a breaking change.

databases/core.py
zanieb: Hm interesting. I'm not sure myself, just wanted to check if you had considered it.

Thanks for chiming in on this after the delay. If you're interested in working on it again I can review and release it. Did you confirm this works and improves performance as expected?

Mng-dev-ai: I can work on it tmr for a few.
it should improve the performance but i didn't make any benchmarks tbh

databases/backends/postgres.py
Mng-dev-ai: yea I don't think it should break anything

PR #470, opened 2022-03-03
Depend on sqlalchemy[asyncio] instead of just sqlalchemy

### LINKED ISSUE #469: Dependency on sqlalchemy should include asyncio extra
Since this library uses async API’s from `sqlalchemy`, it should depend on `sqlalchemy[asyncio]` to ensure all necessary dependencies are installed.

Without this, `ValueError: the greenlet library is required to use this function` occurs on platforms where a `greenlet` binary wheel is not available on PyPI; on those platforms, `sqlalchemy` does not depend on `greenlet`, but its `asyncio` extra does.

See [this Fedora issue](https://bugzilla.redhat.com/show_bug.cgi?id=2012590) for full details and discussion.

I will be happy to provide a PR.

musicinmybrain:
Adding the `asyncio` extra ensures all dependencies for async APIs are required on all platforms.

Fixes #469.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ setup(
     author_email="tom@tomchristie.com",
     packages=get_packages("databases"),
     package_data={"databases": ["py.typed"]},
-    install_requires=["sqlalchemy>=1.4,<1.5", 'aiocontextvars;python_version<"3.7"'],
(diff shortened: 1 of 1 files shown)


musicinmybrain:
This is not needed; see https://github.com/encode/databases/issues/469#issuecomment-1059146540.

PR #473, opened 2022-03-06
allow iterate over custom num of records

### LINKED ISSUE #413: feature request: iterate over more than 1 record

Mng-dev-ai:
Closes #413

### DIFF
--- a/databases/backends/aiopg.py
+++ b/databases/backends/aiopg.py
@@ -179,7 +179,7 @@ class AiopgConnection(ConnectionBackend):
             cursor.close()
 
     async def iterate(
-        self, query: ClauseElement
+        self, query: ClauseElement, n: int = None
     ) -> typing.AsyncGenerator[typing.Any, None]:
         assert self._connection is not None, "Connection is not acquired"
         query_str, args, context = self._compile(query)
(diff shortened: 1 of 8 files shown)


databases/backends/aiopg.py
```diff
@@ -179,7 +179,7 @@ async def execute_many(self, queries: typing.List[ClauseElement]) -> None:
             cursor.close()
 
     async def iterate(
-        self, query: ClauseElement
+        self, query: ClauseElement, n: int = None
```
Euromantic, line 182: `n: Optional[int] = None`

databases/backends/aiopg.py
```diff
@@ -195,6 +195,10 @@ async def iterate(
                     Row._default_key_style,
                     row,
                 )
+                if n is not None:
+                    n -= 1
+                    if n == 0:
```
Euromantic, line 200: Maybe use `if n <= 0` in case if negative `n` was passed

Issue #190, opened 2020-04-22
KeyError when using returning(literal_column('*')) with insert


kamilglod:
slqlalchemy supports `"*"` as a returning value, however when I'm using it with postgres backend 
```python
row = await db.fetch_one(table.insert().values(**values).returning(literal_column('*')))
```
and then try to get row data `KeyError` shows up in this line https://github.com/encode/databases/blob/master/databases/backends/postgres.py#L105
That's because `self._column_map` has value `{'*': (0, NullType())}`

So there should be probably some extra check for the wildcard column.

michael0liver:
I ran into this issue too, this will give you what you want:    
```python
query = table.insert().values(**data).returning(*table.columns)
row = await database.fetch_one(query=query)
```

kamilglod:
Why it is closed? It's a bug, even if it's possible to work around it by explicit passing columns. If this library supports sqlalchemy core queries than it should support asterisk column lookup too.

PR #482, opened 2022-04-01
Fix postgres KeyError with missing explicit columns (#190)

rsommer:
If no explicit colum info is present, the postgres Record class causes a
KeyError. Besides the given example in encode/databases#190 this also
happens if you construct the queries from text-clauses.

Handling this the same way as if there is no entry in `self._column_map`
fixes the error.

A simple testcase is added as well.

I did a simple test run using sqlite and postgres.

### DIFF
--- a/databases/backends/postgres.py
+++ b/databases/backends/postgres.py
@@ -140,7 +140,10 @@ class Record(RecordInterface):
         elif isinstance(key, int):
             idx, datatype = self._column_map_int[key]
         else:
-            idx, datatype = self._column_map[key]
+            try:
+                idx, datatype = self._column_map[key]
+            except KeyError:
+                return self._row[key]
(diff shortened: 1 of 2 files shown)


databases/backends/postgres.py
```diff
@@ -140,7 +140,10 @@ def __getitem__(self, key: typing.Any) -> typing.Any:
         elif isinstance(key, int):
             idx, datatype = self._column_map_int[key]
         else:
-            idx, datatype = self._column_map[key]
+            try:
+                idx, datatype = self._column_map[key]
+            except KeyError:
+                return self._row[key]
```
zanieb, line 146: Should we be setting `raw` to `self._row[key]` to ensure the processor runs on it still?

PR #489, opened 2022-05-17
Adding more to decorator docs

mgjo5899:
I wasn't able to find a way to properly use the transaction decorator when using with other decorators in the docs.
I thought it would be helpful to add this portion so others can also be aware they need to put the decorator right above the function signature without going into the code and finding out about it themselves.

### DIFF
--- a/docs/connections_and_transactions.md
+++ b/docs/connections_and_transactions.md
@@ -95,6 +95,17 @@ async def create_users(request):
     ...
 ```
 
+When using with other decorators (such as route decorator `@router.post("")`
+from FastAPI), try to put `@database.transaction()` right above the function
+signature to ensure the automatic transaction management to be applied:
+
+```python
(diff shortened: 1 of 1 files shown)


docs/connections_and_transactions.md
```diff
@@ -95,6 +95,17 @@ async def create_users(request):
...
 ```
 
+When using with other decorators (such as route decorator `@router.post("")`
+from FastAPI), try to put `@database.transaction()` right above the function
+signature to ensure the automatic transaction management to be applied:
+
+```python
+@router.post("/my-awesome-endpoint")
+@database.transaction()
+async def awesome_endpoint():
```
aminalaee, line 105: I think it would be better to mention something general that "If using multiple decorators, the `@database.transaction()` should be the most inner to the function definition. But maybe with better wording 😄

PR #493, opened 2022-05-31
Improve typing for `Transaction`

### LINKED ISSUE #492: Improve typing
Hi! I've noticed several places 

### `Transaction.__call__`

Right now it is defined as:

```python
    def __call__(self, func: typing.Callable) -> typing.Callable:
        """
        Called if using `@database.transaction()` as a decorator.
        """

        @functools.wraps(func)
        async def wrapper(*args: typing.Any, **kwargs: typing.Any) -> typing.Any:
            async with self:
                return await func(*args, **kwargs)

        return wrapper

```

https://github.com/encode/databases/blob/fbea46d28bda46ba86e940645b456d5ca7f8f5a3/databases/core.py#L356-L366

Which make all wrapped function to have `(*Any, **Any) -> Any` type. This is not ideal.
I propose to use `TypeVar('_CallableType', bound=Callable)` to save the exact type.

After:

```python

db = Database('url')

async def test(a: int) -> int:
    ...

reveal_type(db.transaction()(test))
# note: Revealed type is "def (a: builtins.int) -> typing.Coroutine[Any, Any, builtins.int]"
```

Notice, that signature is preserved.

### `Transaction.__await__`

Right now there's a small problem with it:

```python

db = Database('url')

async def test() -> None:
    reveal_type(await db.transaction())
    # note: Revealed type is "Any"
    # But, should be the same as the next line:
    reveal_type(await db.transaction().start())
    # note: Revealed type is "databases.core.Transaction"
```

I propose to use the same type for `await db.transaction()`

sobolevn:
Closes #492

### DIFF
--- a/databases/core.py
+++ b/databases/core.py
@@ -315,6 +315,9 @@ class Connection:
         return query
 
 
+_CallableType = typing.TypeVar("_CallableType", bound=typing.Callable)
+
+
 class Transaction:
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


Kludex (approved):

aminalaee (approved):
Thank you for the PR 👍

PR #497, opened 2022-07-12
Add Python 3.11 to the pipeline

Kludex:


Kludex:
Dependencies need to be pinned here.

jhominal (approved):

PR #498, opened 2022-07-18
Fix connection leak when transaction commit/rollback raise an exception

masipcat:
This fixes the following case:

```python

import asyncio


async def method():
    database = Database("postgresql://...")
    await database.connect()

    try:
        async with database.connection() as connection:
            await connection.execute(
                """
                CREATE TABLE test (
                    id integer PRIMARY KEY INITIALLY DEFERRED
                );
                """
            )
            async with connection.transaction():
                await connection.execute("insert into test (id) values (1)")
                await connection.execute("insert into test (id) values (1)")
                # The exception UniqueViolationError is raised during txn.commit()
    finally:
        # Without the fix, connection._connection_counter is 1
        print("Before disconnect:", connection._connection_counter)
        # This blocks forever waiting for the leaked connection
        await database.disconnect()
        print("After disconnect")


asyncio.run(method())
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## 0.6.1 (Aug 9th, 2022)
+
+### Fixed
+
+* Improve typing for `Transaction` (#493)
+* Allow string indexing into Record (#501)
+
 ## 0.6.0 (May 29th, 2022)
 
 * Dropped Python 3.6 support (#458)
--- a/databases/__init__.py
+++ b/databases/__init__.py
@@ -1,4 +1,4 @@
 from databases.core import Database, DatabaseURL
 
-__version__ = "0.6.0"
+__version__ = "0.6.1"
 __all__ = ["Database", "DatabaseURL"]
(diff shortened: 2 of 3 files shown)


aminalaee:
Hi, thanks for the PR.
Can you maybe include the tests for it?

masipcat:
Yeah no problem I'll add a test

tests/test_databases.py
```diff
@@ -543,6 +543,33 @@ async def test_transaction_rollback(database_url):
...
+                )
+                async with connection.transaction():
+                    await connection.execute("insert into test (id) values (1)")
+                    await connection.execute("insert into test (id) values (1)")
+
+        # During transaction.commit() postgres will raise this exception:
+        #  asyncpg.exceptions.UniqueViolationError: duplicate key value violates unique constraint "test_pkey"
+        #  DETAIL:  Key (id)=(1) already exists.
+        assert "unique constraint" in str(excinfo.value)
+        assert connection._connection_counter == 0
```
masipcat, line 570: @aminalaee I think this test it's only valid for postgresql. Is there a way to skip for other databases?

aminalaee: Yeah sure, https://github.com/encode/databases/blob/6a52a4a37e69fcb25f309ff5d38e4d43eb80b32c/tests/test_databases.py#L482

tests/test_databases.py
```diff
@@ -575,6 +575,34 @@ async def test_transaction_rollback(database_url):
             assert len(results) == 0
 
 
+@pytest.mark.parametrize("database_url", DATABASE_URLS)
+@mysql_versions
```
aminalaee: If you rebase you don't need `mysql_versions`

PR #501, opened 2022-07-27
Allow string indexing into Record

circlingthesun:
The Record interface inherits from Sequence which only supports integer indexing. The Postgres backend supports string indexing into Records (https://github.com/encode/databases/blob/master/databases/backends/postgres.py#L135).  This PR updates the interface to reflect that.

At least on the Postgres backend __getitem__ deserializes some data types, so it's not equivalent to the _mapping property.

### DIFF
--- a/databases/interfaces.py
+++ b/databases/interfaces.py
@@ -73,3 +73,6 @@ class Record(Sequence):
     @property
     def _mapping(self) -> typing.Mapping:
         raise NotImplementedError()  # pragma: no cover
+
+    def __getitem__(self, key: typing.Union[str | int]) -> typing.Any:
+        raise NotImplementedError()  # pragma: no cover
--- a/databases/interfaces.py
+++ b/databases/interfaces.py
@@ -74,5 +74,5 @@ class Record(Sequence):
     def _mapping(self) -> typing.Mapping:
         raise NotImplementedError()  # pragma: no cover
 
-    def __getitem__(self, key: typing.Union[str | int]) -> typing.Any:
+    def __getitem__(self, key: typing.Any) -> typing.Any:
         raise NotImplementedError()  # pragma: no cover


databases/interfaces.py
```diff
@@ -73,3 +73,6 @@ class Record(Sequence):
     @property
     def _mapping(self) -> typing.Mapping:
         raise NotImplementedError()  # pragma: no cover
+
+    def __getitem__(self, key: typing.Union[str | int]) -> typing.Any:
```
aminalaee, line 77: Hey, should we have this on the interface which all Record classes will inherit?
This is already there in Postgresql backends:
https://github.com/encode/databases/blob/c8770980b9fb6f7d0fbdd3d786f4d2b081572007/databases/backends/postgres.py#L135

circlingthesun: It is indeed. However, all the database query functions like `fetch_one` return type `Record` from interfaces.py, not `Record` from backends/postgres.py. So unless there is a way to setup the `Database` object to return backend specific type information, there is no way to use string indexing with the Postgres backend without MyPy errors all over one's codebase.

circlingthesun: @aminalaee I've done a proof of concept of that approach here: https://github.com/circlingthesun/databases/commit/58dcf587e6b5a4a23821d779e19288148bd3bb78 . The downside is that Python doesn't support default values for generics (https://github.com/python/mypy/issues/3737), so the return type would be `Any` by default. It won't be a breaking change though, and one would simply have to replace `db = Database(...` with `db = Database[postgres.Record](...` to get the database specific record type. Is this approach preferable?

aminalaee: After giving this more thought I agree that the approach offered is not perfect, but it's the simplest and fine for now.

circlingthesun: @aminalaee is that the approach in the PR or the 2nd one? If it's this PR would you mind running the CI again? MyPy raised an issue that's hopefully resolved by the update to the PR. Otherwise I can submit a PR for the latter.

circlingthesun:
@aminalaee is there anything else you need from me?

aminalaee (approved):
Thank you @circlingthesun for the contribution.

PR #507, opened 2022-08-15
Ensure __getattr__ matches __getitem__'s behaviour

circlingthesun:
The `__getitem__` function does extra deserialisation that's not happening in __getattr__. As an example referencing a JSONB field with __getitem__ gets you a `dict` whereas `__getattr__` results in a `str`. This PR aligns their behaviour.

### DIFF
--- a/databases/backends/postgres.py
+++ b/databases/backends/postgres.py
@@ -155,7 +155,7 @@ class Record(RecordInterface):
         return len(self._row)
 
     def __getattr__(self, name: str) -> typing.Any:
-        return self._mapping.get(name)
+        return self.__getitem__(name)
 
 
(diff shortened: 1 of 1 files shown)


JGoutin:
Also, the current `__getattr__` returns `None` when accessing to a non existing attribute instead of raising `AttributeError`, like Python standard behavior, or SQLAlchemy `Row`.

This can be also fixed in this PR by doing the following (Else `KeyError` will be risen):
```python
def __getattr__(self, name: str) -> typing.Any:
    try:
        return self.__getitem__(name)
    except KeyError as e:
        raise AttributeError(e.args[0]) from e
```

PR #510, opened 2022-09-24
Wrap types in `typing.Optional` where applicable

tsunyoku:


Kludex:
Hi @tsunyoku ! :wave: 
Would you mind enabling the flag `https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-optional` on our `setup.cfg`? That way, we can enforce those changes.

tsunyoku:
> Hi @tsunyoku ! 👋 Would you mind enabling the flag https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-optional on our `setup.cfg`? That way, we can enforce those changes.

should be done!

Kludex (approved):
Thanks :pray:

PR #515, opened 2022-10-21
Add documentation about creating tables using sqlalchemy schemas; Close #234

### LINKED ISSUE #234: Documentation lacks guidance on creating database tables
The [Databases Introduction](https://www.encode.io/databases/) executes a raw SQL query like so:

```python
query = """CREATE TABLE HighScores (id INTEGER PRIMARY KEY, name VARCHAR(100), score INTEGER)"""
await database.execute(query=query)
```

The next page, [Database Queries](https://www.encode.io/databases/database_queries/) instead uses SQLAlchemy Core like so:

```python
import sqlalchemy


metadata = sqlalchemy.MetaData()

notes = sqlalchemy.Table(
    "notes",
    metadata,
    sqlalchemy.Column("id", sqlalchemy.Integer, primary_key=True),
    sqlalchemy.Column("text", sqlalchemy.String(length=100)),
    sqlalchemy.Column("completed", sqlalchemy.Boolean),
)

...

database = Database('postgresql://localhost/example')

# Establish the connection pool
await database.connect()

# Execute
query = notes.insert()
await database.execute(query=query, values=values)
```

Nowhere actually shows how to create the tables. **I think the documentation could do a better job at showing a full example that runs and works**.

The documentation does link to [SQLAlchemy core official tutorial](https://docs.sqlalchemy.org/en/latest/core/tutorial.html), however this tutorial talks about "engines" which are nowhere to be found in the `databases` library, so it's not easy to find out how to actually create a table or adapt their examples. One may naively try to use the `database` as the `engine`, which fails with a rather cryptic "AttributeError: 'Database' object has no attribute '_run_visitor'" error.

**I think the documentation could do a better job at explaining how one should map the concepts from the SQLAlchemy tutorial for use in this library**.

Finally, the documentation does not seem to provide a reference section to learn all the methods available and look at their internal documentation without the need for Python's `help()`. **I think the documentation should have a section with the reference documentation for all public types**.

All of this is from the point of view who wants to use `databases` as their only interface to databases, and has no experience with `sqlalchemy`. Given that `sqlalchemy` is the recommended way to use it, I think the library should put some effort into explaining how the both libraries work together.

empicano:
Hi there,

First of all, thank you for your work on this great library.

I had to search for a while in the issues to find out how to create tables based on SQLAlchemy schemas, so I added a section about it in the docs, based on #234. I'm open to all comments 😊 

~ Felix

### DIFF
--- a/docs/database_queries.md
+++ b/docs/database_queries.md
@@ -70,12 +70,49 @@ query = notes.select()
 async for row in database.iterate(query=query):
     ...
 
-# Close all connection in the connection pool
+# Close all connections in the connection pool
 await database.disconnect()
 ```
 
 Connections are managed as task-local state, with driver implementations
 transparently using connection pooling behind the scenes.
+
+## Creating tables
+
+Due to databases wrapping around multiple other packages, you can't use [the usual SQLAlchemy core way to create tables](https://docs.sqlalchemy.org/en/20/core/metadata.html#sqlalchemy.schema.MetaData.create_all). Luckily, there's a neat workaround using schema compilation:
+
+```python
+from databases import Database
+import sqlalchemy
+
+database = Database("postgresql+asyncpg://localhost/example")
+
+# Establish the connection pool
+await database.connect()
+
+metadata = sqlalchemy.MetaData()
+dialect = sqlalchemy.dialects.postgresql.dialect()
+
+# Define your table(s)
+notes = sqlalchemy.Table(
+    "notes",
+    metadata,
+    sqlalchemy.Column("id", sqlalchemy.Integer, primary_key=True),
+    sqlalchemy.Column("text", sqlalchemy.String(length=100)),
+    sqlalchemy.Column("completed", sqlalchemy.Boolean),
+)
+
+# Create tables
+for table in metadata.tables.values():
+    # Set `if_not_exists=False` if you want the query to throw an
+    # exception when the table already exists
+    schema = sqlalchemy.schema.CreateTable(table, if_not_exists=True)
+    query = str(schema.compile(dialect=dialect))
+    await database.execute(query=query)
+
+# Close all connections in the connection pool
+await database.disconnect()
+```
 
 ## Raw queries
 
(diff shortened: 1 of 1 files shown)


empicano:
[The documentation](https://www.encode.io/databases/database_queries/) has a section "Table declarations", followed by the section "Queries", which assumes that those tables now exist.

Personally, I got stuck there because I didn't know how to actually create the tables. I think a note here would be helpful.

I'm not dead set on anything, and I agree that alembic is a more solid choice. I'll add a note about this.

For experiments, and falling back to SQLAlchemy, is there a way to "extract" the engine from databases? I think it would be neat to avoid opening a new connection just to create the tables.

rafalp:
Will the approach you are proposing also handle the constraints and keys? Or will it just create tables?

docs/database_queries.md
```diff
@@ -24,9 +24,48 @@ notes = sqlalchemy.Table(
 )
 ```
 
-You can use any of the sqlalchemy column types such as `sqlalchemy.JSON`, or
+You can use any of the SQLAlchemy column types such as `sqlalchemy.JSON`, or
 custom column types.
 
+## Creating tables
+
+Due to databases wrapping around multiple other packages, you can't use [the usual SQLAlchemy core way to create tables](https://docs.sqlalchemy.org/en/20/core/metadata.html#sqlalchemy.schema.MetaData.create_all). Luckily, there's a neat workaround using schema compilation:
```
rafalp, line 32: We should be more specific why some things are not possible and struggle for neutral language:

```suggestion
Databases don't use SQLAlchemy's engines for database access, making [the usual SQLAlchemy core way to create tables with `create_all`](https://docs.sqlalchemy.org/en/20/core/metadata.html#sqlalchemy.schema.MetaData.create_all) not available. To work around this you can build `TABLE CREATE` queries manually:
```

docs/database_queries.md
```diff
@@ -24,9 +24,48 @@ notes = sqlalchemy.Table(
...
+    # exception when the table already exists
+    schema = sqlalchemy.schema.CreateTable(table, if_not_exists=True)
+    query = str(schema.compile(dialect=dialect))
+    await database.execute(query=query)
+
+# Close all connections in the connection pool
+await database.disconnect()
+```
+
+Note that this way of creating tables is only useful for experiments. In a production environment, you'd rather use something like [Alembic](https://alembic.sqlalchemy.org/en/latest/), which also helps with database migrations.
```
rafalp, line 67:
```suggestion
Note that this way of creating tables is only useful for local experimentation. For serious projects, we recommend using proper database schema migrations solution like [Alembic](https://alembic.sqlalchemy.org/en/latest/)
```

rafalp (changes requested):
Wording suggestions.

rafalp:
Left some notes. Sorry for taking days of delay with those, I have plenty of projects on my plate and trying to manage as much as I can until fulltime maintainer steps up for databases 🙏

empicano:
No worries about the delay!

I added your wording suggestions and fixed some typos 👍 

> Will the approach you are proposing also handle the constraints, keys and indexes? Or will it just create tables?

Yes. Basically, we define the SQLAlchemy schema, then compile it to a raw SQL query, then execute it with databases. The example I'm proposing to add to the documentation compiles to this (note the primary key constraint):

```
CREATE TABLE IF NOT EXISTS notes (
    id SERIAL NOT NULL, 
    text VARCHAR(100), 
    completed BOOLEAN, 
    PRIMARY KEY (id)
)
```

This compilation is a [standard part of SQLAlchemy](https://docs.sqlalchemy.org/en/20/faq/sqlexpressions.html#how-do-i-render-sql-expressions-as-strings-possibly-with-bound-parameters-inlined). When we define the database dialect (PostgreSQL, SQLite, ...) that we want to compile to -- as we do in the example -- I don't see that there are any limitations.

rafalp (approved):

rafalp:
Looking good to me, thanks!

PR #517, opened 2022-10-24
docs: Update sqlalchemy core tutorial link

njanke96:
The current link in the docs (specifying the version "latest") now points to the docs for the SQLAlchemy 2.0 beta. This PR changes the link to strictly specify version 1.4.

It would also be possible to change the link to point to "stable" rather than "latest", however 1.4 seemed more appropriate as "stable" will likely point to 2.0 before this library migrates.

Cheers


rafalp (approved):
I agree with this change until SQLAlchemy 2.0 stabilises and library is updated to support it.

Can you please update on master so I can merge it? Thanks!

njanke96:
Updated

rafalp:
Thank you!

PR #519, opened 2022-11-03
add support rowcount for asyncpg

nicolaichuk:
Please review and merge to upstream.

### DIFF
--- a/databases/backends/postgres.py
+++ b/databases/backends/postgres.py
@@ -214,7 +214,29 @@ class PostgresConnection(ConnectionBackend):
     async def execute(self, query: ClauseElement) -> typing.Any:
         assert self._connection is not None, "Connection is not acquired"
         query_str, args, result_columns = self._compile(query)
-        return await self._connection.fetchval(query_str, *args)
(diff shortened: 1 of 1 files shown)


rafalp:
Thank you for contributing. Can you update us on what problem those changes solve or what feature they add?

nicolaichuk:
> Thank you for contributing. Can you update us on what problem those changes solve or what feature they add?

@rafalp  This PR add in postgres backends  the ability to get a new value of primary key when a new row is inserted, and the number of rows changed when performing other operations.

You can see the implementation of this behavior for the library here sqlite backends https://github.com/encode/databases/blob/0.7.0/databases/backends/sqlite.py#L134-L136

Please review and merge to upstream.

rafalp (approved):

rafalp (changes requested):
Those changes look okay, but we should have tests to make sure they work and this behavior doesn't break in future by other PR.

zanieb:
Closing this as stale. Feel free to open again if you want to add tests.

PR #539, opened 2023-03-21
Integrating SQLALchemy 2.0

tarsil:
## Attention

- This introduces the minimum requirement of python 3.8 (3.7 EOL support is in June anyway)
- Removed support for aiopg as it does not support SQLA 2.0 (so far). It can be added back later if they decide to add it.
- Added fix for the _CompileLabel introduced as annotations for SQLA 2.0
- Other minor fixes and updates in the docs
- Removed tests that are no longer required.
- Updated tests to remove references to aiopg

@Kludex and @aminalaee  - I didn't have a lot of time recently and this, at least locally, was passing the current tests


tarsil:
Weird, locally it was passing but I was using Postgres only to be honest. Hmm, it would be great to have some suggestions. As I mentioned before, didn't have too much time on this, unfortunately.

databases/__init__.py
```diff
@@ -1,4 +1,4 @@
 from databases.core import Database, DatabaseURL
 
-__version__ = "0.7.0"
+__version__ = "0.8.0"
```
Kludex, line 3:
```suggestion
__version__ = "0.7.0"
```

CHANGELOG.md
```diff
@@ -1,126 +1,4 @@
 # Changelog
 
-All notable changes to this project will be documented in this file.
+[Visit](https://www.encode.io/databases/release-notes) to see the changelog.
 
-The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
```
aminalaee, line 5: I think the PR is getting way too big, for no reason. can we please stick only to necessary changes?
For example I think CHANGELOG.md is consistent withing all encode projects.

aminalaee (commented):
Thank you for the effort, just a general note.

tarsil:
@Kludex and @aminalaee I can put back the changelog to the way it was. I just thought it was to be consistent with Starlette and other products. Usually when people go after docs, having the release notes there helps. I'm just fixing some of the issues with the CI and this should be ok.

I was able to reintegrate with aiopg as well and the changes themselves not only cleans but also updates a lot of things based on SQLAlchemy 2.0

tarsil:
I will check what is happening here with the CI and then I will close this PR and open a clean new one @aminalaee and @Kludex

tarsil:
Opening a brand new and cleaner PR

PR #540, opened 2023-03-22
🪛 Moving to SQLAlchemy 2.0

tarsil:
* Added support for python 3.11
* Added common and dialects packages to handle the new SQLAlchemy 2.0+

@aminalaee @Kludex this is a cleaner version of the previous PR with the changes passing the new SQLAlchemy 2.0 which is significantly different from the < 2.0

It is also good to mention that when I was doing this PR, I was reading some suggestions and the one from here https://github.com/encode/databases/pull/507 was good to add it as well. So I would like to thank @circlingthesun  and @JGoutin for the suggestions

Edit: The asyncpg sometimes hangs during the tests. I forcibly close the connections during that test to make sure it doesn't get stuck

PR #544, opened 2023-03-28
Support for SQL Server

tarsil:
@aminalaee @Kludex 

This is a PR proposing support for SQL Server. A lot of companies still use it and with `aioodbc` and databases patterns, this was possible to achieve as well. This could be merged maybe after [this](https://github.com/encode/databases/pull/540) one? The reason why after its only because I branched off that branch to create this new one and once it's merged, it is way less files to review.


rafalp:
Do we want to extend Databases with new backends when library is basically on life support? 🤔

tarsil:
@rafalp the lib is actually very stable and works for 99% of the projects as is. With the latest 2PR being reviewed we go to the next level. Unless @aminalaee and @Kludex feel that Encode doesn't want to continue with this. If that is the case, there is always a way to continue anyway :) but I believe Encode wants to keep maintaining this :)

rafalp:
> I believe Encode wants to keep maintaining this

I'm also Encode member and I was on meetings where we've discussed how to carry on with databases with my proposition being to keep it around because its great as lightweight async DBAL alternative to solve-it-all SQlAlchemy 2 ;)

tarsil:
@rafalp I didn't mean any disrespect about you ahahah. I respect Encode tremendously :D. A lot of what I do is inspired by the great work you all do.

I have more than interest to use databases and the reason why I "created" the driver for SQL Server is because I'm currently working on projects that use heavily SQL Server and the amount of code being used there is so ugly that databases solves it in a simple fashion. I never wanted to fork databases in order to put SQLAlchemy 2 and that is why I tried to open the PR [here](https://github.com/encode/databases/pull/540) with the migration working but I also believe Encode knows what is the best course of action and I will always support the decisions 👍🏼 

If you are Encode, I'm sure you are a fantastic developer! :D 

This PR althoug it looks big, it is mostly because I branched of the one implementing SQLAlchemy 2.0 but not yet merged

rafalp:
TBH I've always been thinking that Databases should support external drivers, so adding new databases to it should not changing the core library.

I'm not familiar with SQL Server but this PR seems to mix SQLAlchemy 2 support changes with SQL Server changes.

tarsil:
> TBH I've always been thinking that Databases should support external drivers, so adding new databases to it should not changing the core library.
> 
> I'm not familiar with SQL Server but this PR seems to mix SQLAlchemy 2 support changes with SQL Server changes.

No, this is the same level of integration that is done for Postgres and MySQL but supporting SQLServer as well. SQL Alchemy has support for it already. What I'm proposing is simply to also add the same layer to MSSQL since it's widely used.

The PR it looks big beause the SQL Alchemy 2.0 changes **were not yet merged**. It is [here](https://github.com/encode/databases/pull/540). Because I wanted to use already SQLA 2.0, I branched off that branch and added the SQLServer driver. Once that PR is merged, if is merged, you should only see 3 or 4 files **only* here to review

Issue #456, opened 2022-01-27
Connections and transactions meta issue


yeraydiazdiaz:
Hi all, we've been experiencing some issues with Databases regarding connection and transaction management, so I've spent some time reading through the different issues and PRs and understanding the code. There are couple cross cutting problems that I think need discussion as to how to best tackle them. I'll do my best to outline the past and pressent issues and their causes.

1. Databases makes use of [ContextVars](https://docs.python.org/3/library/contextvars.html) to store the Connection objects across tasks. When a query method (execute, fetch, etc) is awaited the `connection` method is called which checks for the presence of a connection in the context, if there is one it is returned, otherwise it creates a new connection and sets it in the context (https://github.com/encode/databases/blob/master/databases/core.py#L190-L202).
2. However, as outlined in https://github.com/encode/databases/issues/230, once a connection is created all coroutines/tasks _inherit the context from the parent_, which effectively means the same connection will be returned. It's unclear to me if this was an intentional decision or an oversight, but it's the cause of issues regarding transactions opened in concurrent tasks, reported in https://github.com/encode/databases/issues/340 and https://github.com/encode/databases/issues/327.
3. A [PR](https://github.com/encode/databases/pull/328) was opened and merged addressing this issue which essentially moves away from the previous behaviour of one connection being shared to each time `transaction` is called [a new connection is created](https://github.com/encode/databases/blob/master/databases/core.py#L208-L214). This change was part of the [0.5.0 release](https://github.com/encode/databases/blob/master/CHANGELOG.md#050-august-26th-2021). However, this change led to issues where transactions are not rolled back correctly (https://github.com/encode/databases/issues/403, https://github.com/encode/databases/issues/424, https://github.com/encode/databases/issues/425), although it's not entirely clear to me why (any clarification on this respect would be appreciated).

In summary, there's two valid use cases, concurrent transaction management and predictable non-concurrent transaction rollback, that seem at odds with each other in the current implementation. Ideally we'd find a solution for https://github.com/encode/databases/issues/327 that does not require creating a new connection per transaction.

I also think it might be worth clarifying the connection management model, the [docs](https://www.encode.io/databases/connections_and_transactions/) suggest the connections are handled transparently but it's unclear when actual backend connections are created, from the code it seems they're [only created once on `Connection.__aenter__`](https://github.com/encode/databases/blob/master/databases/core.py#L248-L257) and since the first created connection is set in the context and returned at all times, there's effectively a single backend connection acquired at any point in time.

[...]

Issue #340, opened 2021-05-16
PostgreSQL: Transaction rollback errors with "another operation is in progress"


rafalp:
I've noticed that if one of my tests fails, every next test will fail with error from `async with database` claiming that connection is still on.

Upon closer look I've found out that first tests failure triggers transaction's rollback which sends `ROOLBACK START` to the database.

However `asyncpg.exceptions._base.InterfaceError: cannot perform operation: another operation is in progress` will be raised in response to this, resulting in invalid state:

```
is_connected = True
_transaction_stack = []
```

This state can't be fixed other by monkeypatching connection object, as trying to call `disconnect()` will try to pop item from empty `_transaction_stack`, causing other exception.

Min repro:

```
database = Database(url="...", force_rollback=True)

async with database:
    raise ValueError()

async with database:  # this will crash because previous database connection was not cleaned up completely.
   database.exec("...")
```

Issue #134, opened 2019-08-09
Errors with multiple tasks and transactions in postgresql and sqlite


reclosedev:
Minimal reproduce script with comments

```python
import asyncio

import databases
import sqlalchemy as sa

# import logging
# logging.basicConfig(level=logging.DEBUG)


DB_URL = "sqlite:////tmp/test.db"

metadata = sa.MetaData()
table = sa.Table(
    "t1", metadata,
    sa.Column("id", sa.Integer()),
    sa.Column("data", sa.String()),
)
database = databases.Database(DB_URL)


async def init_db():
    engine = sa.create_engine(DB_URL)
    metadata.create_all(engine)
    await database.connect()


async def work(start: int, n=5, data="a"):
    async with database.transaction():
        query = table.insert()
        for i in range(start, n):
            await database.execute(query, {"id": i, "data": data})
    print(f"Ok {start}")


async def main_async():
    await init_db()
    futures = []
    some_query = table.select()
    # Error is raised if there is query executed before creating futures
    # Doesn't matter if transaction is explicit or not
    async with database.transaction():
        # !!! If I remove this query (from parent task), it works fine
        await database.fetch_all(some_query)
    step = 5
    for i in range(0, 4 * step, step):
        fut = asyncio.ensure_future(work(i, i + step))
        futures.append(fut)
    await asyncio.wait(futures)
    print("Done")


loop = asyncio.get_event_loop()
loop.run_until_complete(main_async())

```
Error with sqlite database: 
```
databases/core.py", line 305, in commit
assert self._connection._transaction_stack[-1] is self
```
Error with postgresql database:
```
  File "asyncpg/protocol/protocol.pyx", line 301, in query
  File "asyncpg/protocol/protocol.pyx", line 659, in asyncpg.protocol.protocol.BaseProtocol._check_state
asyncpg.exceptions._base.InterfaceError: cannot perform operation: another operation is in progress
```

reclosedev:
Looks like the root cause of this issue is how `contexvars` works. I've tried 3.6 aiocontextvars and 3.7 standart's library verision.

If context var is set in outer scope, then all inner tasks have the same value, so new connection is not created in new tasks.

If you change code in example to
```python
    new_context = contextvars.Context()
    for i in range(0, 4 * step, step):
        fut = new_context.run(asyncio.ensure_future, work(i, i + 5))
        futures.append(fut)
    await asyncio.wait(futures)
```
then it works as expected. But I don't think that it's a good solution.

[...]

Issue #123, opened 2019-07-09
Assertion error when using transaction decorator


SebasAren:
I'm using fastapi in combination with this library but when I use the transaction decorator I'm sporadically getting the following assertion error: 
```
File ".../lib/python3.7/site-packages/databases/core.py", line 305, in commit
    assert self._connection._transaction_stack[-1] is self
```

I'm using a pretty basic fastapi setup making use of the on event startup hook to connect the database. The database is mysql, which i'm connecting to with the aiomysql package (and this library).

I've decorated all of my routes in the following manner :
```
@app.get("/", response_model=DataModel)
@database.transaction()
async def hello_world(q: str):
    query = sqlalchemy.select(table)
    result = await database.fetch_all(query)
    return result
```
This works most of the time, but once in a while it raises the above exception. If I remove the transaction decorator everything works as expected. 

Why do I get this error? Am I using the decorator improperly?

tomchristie:
Not sure - could be to do with the interfaction between the two decorators maybe?
Have you tried them the other way around?
Are there any issues on FastAPI's GitHub repo related to this?
Are there example examples in FastAPI of using it together with `databases`?

SebasAren:
> Not sure - could be to do with the interfaction between the two decorators maybe?

This has crossed my mind and seems the most logical explanation

> Have you tried them the other way around?

That seems to me to be the "wrong" way around. But I might try it out.

> Are there any issues on FastAPI's GitHub repo related to this?

Not that I could find.  

> Are there example examples in FastAPI of using it together with `databases`?

The main fastapi documention has an example to set it up, but doesn't use transactions and is pretty basic overall.

While typing this I got another idea of how I could set this up. Using a middleware to attach a transaction to each route, which sounds a lot less error prone. I'm afraid the problem is because of the double decorator.

circlingthesun:
This took me a while to figure out. I used locust to create some load tests. This issue only happens when two concurrent requests happen on the same endpoint. Calling database.transaction() creates a Transaction object which is then used to decorate a route function. This instance of the Transaction object's state is then shared between all invocations of the route function.

[...]

PR #546, opened 2023-04-07
fix: incorrect concurrent usage of connection and transaction

zevisert:
> [!NOTE]
> EDIT: This PR contains some back and forth on how _**databases**_ should handle concurrent connections across tasks. If you're looking into _**databases**_ concurrency model, it's worth taking a read through all of our comments. The specific issue with the `@Database.transaction` decorator was fixed, but it required digging into _and changing_ how connection state is tracked across `asyncio.Tasks` using `asyncio.ContextVars`

---

This commit should fix an issue with shared state with transactions.

The main issue is very well described by @circlingthesun here: https://github.com/encode/databases/issues/123#issuecomment-1028486810 - we also ran into this most reliably using locust to load test as well, but did find this issue a few times in normal usage. 

The big problem is that using the `@db.transaction` decorator creates a new `Databases.Transaction` instance that, when called creates a context manager for a transaction. The problem is that in concurrent frameworks, such as ASGI servers like starlette / fastapi, it's possible to have two concurrent tasks running. Since the Transaction instance is shared between tasks, it's `self` variable is comparable to shared global state, concurrent code can't trust that no race condition has occurred without some helper mechanism like a ContextVar.

Here's a mermaid diagram trying to show this: 
```mermaid
sequenceDiagram
    actor User1
    actor User2
    participant API
    participant TXN
    API->>TXN: (startup) @db.transaction creates new shared Transaction(...) for route
    User1->>API: (1): Request to API route creates new asyncio.Task(...)
    API->>TXN: (1): Transaction.__call__ assignes self._connection
    User2->>API: (2): Request to same API route creates new asyncio.Task(...)
    API->>TXN: (2): PROBLEM -- Transaction.__call__ obtains new Databases.Connection for this asyncio.Task,<br/> like above, but overwrites self._connection
    TXN->>API: (1): PROBLEM -- Transaction.__aexit__ uses self._connection of task (2)<br/>since it was not retrieved from a ContextVar
    API->>User1: (1): Responding here closes the transaction's _connection
    TXN->>API: (2): FAULT -- Transaction.__aexit__ expects to also close self._connection, <br/>but since it also tries to retrieve the connection from self._connection instead of the ContextVar,<br/>it retireves (1)'s already closed connection
    API->>User2: (2): FAULT -- Propigated exception results in 500 response code
```

All I've done with this PR is move the shared global state of `self._connection` and `self._transaction` into local or contextvars instead of instance variables.

Related issues are: #123 #134
Possibly related: #340
Related meta issue: #456

### DIFF
--- a/databases/core.py
+++ b/databases/core.py
@@ -11,7 +11,7 @@ from sqlalchemy import text
 from sqlalchemy.sql import ClauseElement
 
 from databases.importer import import_from_string
-from databases.interfaces import DatabaseBackend, Record
+from databases.interfaces import DatabaseBackend, Record, TransactionBackend
 
 try:  # pragma: no cover
     import click
@@ -344,6 +344,9 @@ class Transaction:
         self._connection_callable = connection_callable
         self._force_rollback = force_rollback
         self._extra_options = kwargs
+        self._transaction_context: ContextVar[TransactionBackend | None] = ContextVar(
+            "transaction_context"
+        )
 
     async def __aenter__(self) -> "Transaction":
         """
@@ -385,31 +388,38 @@ class Transaction:
         return wrapper  # type: ignore
 
     async def start(self) -> "Transaction":
-        self._connection = self._connection_callable()
-        self._transaction = self._connection._connection.transaction()
+        connection = self._connection_callable()
+        transaction = connection._connection.transaction()
+        self._transaction_context.set(transaction)
 
-        async with self._connection._transaction_lock:
-            is_root = not self._connection._transaction_stack
-            await self._connection.__aenter__()
-            await self._transaction.start(
-                is_root=is_root, extra_options=self._extra_options
-            )
-            self._connection._transaction_stack.append(self)
+        async with connection._transaction_lock:
+            is_root = not connection._transaction_stack
+            await connection.__aenter__()
+            await transaction.start(is_root=is_root, extra_options=self._extra_options)
+            connection._transaction_stack.append(self)
         return self
 
     async def commit(self) -> None:
-        async with self._connection._transaction_lock:
-            assert self._connection._transaction_stack[-1] is self
-            self._connection._transaction_stack.pop()
-            await self._transaction.commit()
-            await self._connection.__aexit__()
+        connection = self._connection_callable()
+        transaction = self._transaction_context.get()
+        assert transaction is not None, "Transaction not found in current task"
+        async with connection._transaction_lock:
+            assert connection._transaction_stack[-1] is self
+            connection._transaction_stack.pop()
+            await transaction.commit()
+            await connection.__aexit__()
+            self._transaction_context.set(None)
 
     async def rollback(self) -> None:
-        async with self._connection._transaction_lock:
-            assert self._connection._transaction_stack[-1] is self
-            self._connection._transaction_stack.pop()
-            await self._transaction.rollback()
-            await self._connection.__aexit__()
+        connection = self._connection_callable()
+        transaction = self._transaction_context.get()
+        assert transaction is not None, "Transaction not found in current task"
+        async with connection._transaction_lock:
+            assert connection._transaction_stack[-1] is self
+            connection._transaction_stack.pop()
+            await transaction.rollback()
+            await connection.__aexit__()
+            self._transaction_context.set(None)
 
 
 class _EmptyNetloc(str):
--- a/databases/core.py
+++ b/databases/core.py
@@ -193,7 +193,7 @@ class Database:
         if self._global_connection is not None:
             return self._global_connection
 
-        connection = self._connection_contextvar.get(default=None)
+        connection = self._connection_contextvar.get(None)
         if connection is None:
             connection = Connection(self._backend)
             self._connection_contextvar.set(connection)
@@ -417,7 +417,7 @@ class Transaction:
 
     async def commit(self) -> None:
         connection = self._connection_callable()
-        transaction = self._transaction_contextvar.get(default=None)
+        transaction = self._transaction_contextvar.get(None)
         assert transaction is not None, "Transaction not found in current task"
         async with connection._transaction_lock:
             assert connection._transaction_stack[-1] is self
@@ -429,7 +429,7 @@ class Transaction:
 
     async def rollback(self) -> None:
         connection = self._connection_callable()
-        transaction = self._transaction_contextvar.get(default=None)
+        transaction = self._transaction_contextvar.get(None)
         assert transaction is not None, "Transaction not found in current task"
         async with connection._transaction_lock:
             assert connection._transaction_stack[-1] is self
--- a/databases/core.py
+++ b/databases/core.py
@@ -36,8 +36,8 @@ except ImportError:  # pragma: no cover
 logger = logging.getLogger("databases")
 
 # Connections are stored as task-local state, but care must be taken to ensure
-# that two database instances in the same task overwrite each other's connections.
-# For this reason, key comprises the database instance and the current task.
+# that two database instances in the same task do not overwrite each other's connections.
+# For this reason, the dict key comprises the database instance and the current task.
 _connection_contextmap: ContextVar[
     Dict[Tuple["Database", asyncio.Task], "Connection"]
 ] = ContextVar("databases:Connection")
--- a/databases/core.py
+++ b/databases/core.py
@@ -5,7 +5,7 @@ import logging
 import typing
 from contextvars import ContextVar
 from types import TracebackType
-from typing import Dict, Optional
+from typing import Dict
 from urllib.parse import SplitResult, parse_qsl, unquote, urlsplit
 
 from sqlalchemy import text
@@ -127,6 +127,7 @@ class Database:
             self._global_connection = None
         else:
             task = asyncio.current_task()
+            assert task is not None, "Not running in an asyncio task"
             connections = _get_connection_contextmap()
             if (self, task) in connections:
                 del connections[self, task]
@@ -204,6 +205,7 @@ class Database:
             return self._global_connection
 
         task = asyncio.current_task()
+        assert task is not None, "Not running in an asyncio task"
         connections = _get_connection_contextmap()
         if (self, task) not in connections:
             connections[self, task] = Connection(self._backend)
--- a/databases/core.py
+++ b/databases/core.py
@@ -5,7 +5,7 @@ import logging
 import typing
 from contextvars import ContextVar
 from types import TracebackType
-from typing import Optional
+from typing import Dict, Optional
 from urllib.parse import SplitResult, parse_qsl, unquote, urlsplit
 
 from sqlalchemy import text
@@ -35,6 +35,21 @@ except ImportError:  # pragma: no cover
 
 logger = logging.getLogger("databases")
 
+# Connections are stored as task-local state, but care must be taken to ensure
+# that two database instances in the same task overwrite each other's connections.
+# For this reason, key comprises the database instance and the current task.
+_connection_contextmap: ContextVar[
+    Dict[tuple["Database", asyncio.Task], "Connection"]
+] = ContextVar("databases:Connection")
+
+
+def _get_connection_contextmap() -> Dict[tuple["Database", asyncio.Task], "Connection"]:
+    connections = _connection_contextmap.get(None)
+    if connections is None:
+        connections = {}
+        _connection_contextmap.set(connections)
+    return connections
+
 
 class Database:
     SUPPORTED_BACKENDS = {
@@ -63,14 +78,6 @@ class Database:
         backend_cls = import_from_string(backend_str)
         assert issubclass(backend_cls, DatabaseBackend)
         self._backend = backend_cls(self.url, **self.options)
-
-        # Connections are stored as task-local state, and cannot be garbage collected,
-        # since the immutable global Context stores a strong reference to each ContextVar
-        # that is created. We need these local ContextVars since two Database objects
-        # could run in the same asyncio.Task with connections to different databases.
-        self._connection_contextvar: ContextVar[Optional["Connection"]] = ContextVar(
-            f"databases:Database:{id(self)}"
-        )
 
         # When `force_rollback=True` is used, we use a single global
         # connection, within a transaction that always rolls back.
@@ -119,7 +126,10 @@ class Database:
             self._global_transaction = None
             self._global_connection = None
         else:
-            self._connection_contextvar.set(None)
+            task = asyncio.current_task()
+            connections = _get_connection_contextmap()
+            if (self, task) in connections:
+                del connections[self, task]
 
         await self._backend.disconnect()
         logger.info(
@@ -193,12 +203,12 @@ class Database:
         if self._global_connection is not None:
             return self._global_connection
 
-        connection = self._connection_contextvar.get(None)
-        if connection is None:
-            connection = Connection(self._backend)
-            self._connection_contextvar.set(connection)
+        task = asyncio.current_task()
+        connections = _get_connection_contextmap()
+        if (self, task) not in connections:
+            connections[self, task] = Connection(self._backend)
 
-        return connection
+        return connections[self, task]
 
     def transaction(
         self, *, force_rollback: bool = False, **kwargs: typing.Any
@@ -339,6 +349,19 @@ class Connection:
 
 _CallableType = typing.TypeVar("_CallableType", bound=typing.Callable)
 
+_transaction_contextmap: ContextVar[
+    Dict["Transaction", TransactionBackend]
+] = ContextVar("databases:Transactions")
+
+
+def _get_transaction_contextmap() -> Dict["Transaction", TransactionBackend]:
+    transactions = _transaction_contextmap.get(None)
+    if transactions is None:
+        transactions = {}
+        _transaction_contextmap.set(transactions)
+
+    return transactions
+
 
 class Transaction:
     def __init__(
@@ -350,15 +373,6 @@ class Transaction:
         self._connection_callable = connection_callable
         self._force_rollback = force_rollback
         self._extra_options = kwargs
-
-        # This ContextVar can never be garbage collected - similar to the ContextVar
-        # at Database._connection_contextvar - since the current Context has a strong
-        # reference to every ContextVar that is created. We need local ContextVars since
-        # there may be multiple (even nested) transactions in a single asyncio.Task,
-        # which each need their own unique TransactionBackend object.
-        self._transaction_contextvar: ContextVar[
-            Optional[TransactionBackend]
-        ] = ContextVar(f"databases:Transaction:{id(self)}")
 
     async def __aenter__(self) -> "Transaction":
         """
@@ -402,12 +416,8 @@ class Transaction:
     async def start(self) -> "Transaction":
         connection = self._connection_callable()
         transaction = connection._connection.transaction()
-
-        # Cannot store returned reset token anywhere, for the same reason
-        # we need a ContextVar in the first place - `self` is not
-        # a safe object on which to store references for concurrent code.
-        self._transaction_contextvar.set(transaction)
-
+        transactions = _get_transaction_contextmap()
+        transactions[self] = transaction
         async with connection._transaction_lock:
             is_root = not connection._transaction_stack
             await connection.__aenter__()
@@ -417,27 +427,27 @@ class Transaction:
 
     async def commit(self) -> None:
         connection = self._connection_callable()
-        transaction = self._transaction_contextvar.get(None)
+        transactions = _get_transaction_contextmap()
+        transaction = transactions.get(self, None)
         assert transaction is not None, "Transaction not found in current task"
         async with connection._transaction_lock:
             assert connection._transaction_stack[-1] is self
             connection._transaction_stack.pop()
             await transaction.commit()
             await connection.__aexit__()
-            # Have no reset token, set to None instead
-            self._transaction_contextvar.set(None)
+            del transactions[self]
 
     async def rollback(self) -> None:
         connection = self._connection_callable()
-        transaction = self._transaction_contextvar.get(None)
+        transactions = _get_transaction_contextmap()
+        transaction = transactions.get(self, None)
         assert transaction is not None, "Transaction not found in current task"
         async with connection._transaction_lock:
             assert connection._transaction_stack[-1] is self
             connection._transaction_stack.pop()
             await transaction.rollback()
             await connection.__aexit__()
-            # Have no reset token, set to None instead
-            self._transaction_contextvar.set(None)
+            del transactions[self]
 
 
 class _EmptyNetloc(str):
--- a/databases/core.py
+++ b/databases/core.py
@@ -11,7 +11,7 @@ from sqlalchemy import text
 from sqlalchemy.sql import ClauseElement
 
 from databases.importer import import_from_string
-from databases.interfaces import DatabaseBackend, Record
+from databases.interfaces import DatabaseBackend, Record, TransactionBackend
 
 try:  # pragma: no cover
     import click
@@ -63,8 +63,8 @@ class Database:
         assert issubclass(backend_cls, DatabaseBackend)
         self._backend = backend_cls(self.url, **self.options)
 
-        # Connections are stored as task-local state.
-        self._connection_context: ContextVar = ContextVar("connection_context")
+        # Connections are stored per asyncio task
+        self._connections: typing.Dict[typing.Optional[asyncio.Task], Connection]= {}
 
         # When `force_rollback=True` is used, we use a single global
         # connection, within a transaction that always rolls back.
@@ -113,7 +113,7 @@ class Database:
             self._global_transaction = None
             self._global_connection = None
         else:
-            self._connection_context = ContextVar("connection_context")
+            self._connections.pop(asyncio.current_task(), None)
 
         await self._backend.disconnect()
         logger.info(
@@ -187,12 +187,12 @@ class Database:
         if self._global_connection is not None:
             return self._global_connection
 
-        try:
-            return self._connection_context.get()
-        except LookupError:
+        current_task = asyncio.current_task()
+        if current_task not in self._connections:
             connection = Connection(self._backend)
-            self._connection_context.set(connection)
-            return connection
+            self._connections[current_task] = connection
+
+        return self._connections[current_task]
 
     def transaction(
         self, *, force_rollback: bool = False, **kwargs: typing.Any
@@ -344,6 +344,9 @@ class Transaction:
         self._connection_callable = connection_callable
         self._force_rollback = force_rollback
         self._extra_options = kwargs
+        
+        # Transactions are stored per asyncio task
+        self._transactions: typing.Dict[typing.Optional[asyncio.Task], "TransactionBackend"]= {}
 
     async def __aenter__(self) -> "Transaction":
         """
@@ -385,31 +388,35 @@ class Transaction:
         return wrapper  # type: ignore
 
     async def start(self) -> "Transaction":
-        self._connection = self._connection_callable()
-        self._transaction = self._connection._connection.transaction()
+        connection = self._connection_callable()
+        transaction = self._transactions[asyncio.current_task()] = connection._connection.transaction()
 
-        async with self._connection._transaction_lock:
-            is_root = not self._connection._transaction_stack
-            await self._connection.__aenter__()
-            await self._transaction.start(
+        async with connection._transaction_lock:
+            is_root = not connection._transaction_stack
+            await connection.__aenter__()
+            await transaction.start(
                 is_root=is_root, extra_options=self._extra_options
             )
-            self._connection._transaction_stack.append(self)
+            connection._transaction_stack.append(self)
         return self
 
     async def commit(self) -> None:
-        async with self._connection._transaction_lock:
-            assert self._connection._transaction_stack[-1] is self
-            self._connection._transaction_stack.pop()
-            await self._transaction.commit()
-            await self._connection.__aexit__()
+        connection = self._connection_callable()
+        transaction = self._transactions[asyncio.current_task()]
+        async with connection._transaction_lock:
+            assert connection._transaction_stack[-1] is self
+            connection._transaction_stack.pop()
+            await transaction.commit()
+            await connection.__aexit__()
 
     async def rollback(self) -> None:
-        async with self._connection._transaction_lock:
-            assert self._connection._transaction_stack[-1] is self
-            self._connection._transaction_stack.pop()
-            await self._transaction.rollback()
-            await self._connection.__aexit__()
+        connection = self._connection_callable()
+        transaction = self._transactions[asyncio.current_task()]
+        async with connection._transaction_lock:
+            assert connection._transaction_stack[-1] is self
+            connection._transaction_stack.pop()
+            await transaction.rollback()
+            await connection.__aexit__()
 
 
 class _EmptyNetloc(str):
--- a/databases/core.py
+++ b/databases/core.py
@@ -5,7 +5,7 @@ import logging
 import typing
 from contextvars import ContextVar
 from types import TracebackType
-from typing import Dict
+from typing import Dict, Tuple
 from urllib.parse import SplitResult, parse_qsl, unquote, urlsplit
 
 from sqlalchemy import text
@@ -39,11 +39,11 @@ logger = logging.getLogger("databases")
 # that two database instances in the same task overwrite each other's connections.
 # For this reason, key comprises the database instance and the current task.
 _connection_contextmap: ContextVar[
-    Dict[tuple["Database", asyncio.Task], "Connection"]
+    Dict[Tuple["Database", asyncio.Task], "Connection"]
 ] = ContextVar("databases:Connection")
 
 
-def _get_connection_contextmap() -> Dict[tuple["Database", asyncio.Task], "Connection"]:
+def _get_connection_contextmap() -> Dict[Tuple["Database", asyncio.Task], "Connection"]:
     connections = _connection_contextmap.get(None)
     if connections is None:
         connections = {}
--- a/databases/core.py
+++ b/databases/core.py
@@ -63,7 +63,7 @@ class Database:
         self._backend = backend_cls(self.url, **self.options)
 
         # Connections are stored per asyncio task
-        self._connections: typing.Dict[typing.Optional[asyncio.Task], Connection] = {}
+        self._connections: typing.Dict[asyncio.Task, Connection] = {}
 
         # When `force_rollback=True` is used, we use a single global
         # connection, within a transaction that always rolls back.
--- a/databases/core.py
+++ b/databases/core.py
@@ -5,6 +5,7 @@ import logging
 import typing
 from contextvars import ContextVar
 from types import TracebackType
+from typing import Optional
 from urllib.parse import SplitResult, parse_qsl, unquote, urlsplit
 
 from sqlalchemy import text
@@ -63,8 +64,13 @@ class Database:
         assert issubclass(backend_cls, DatabaseBackend)
         self._backend = backend_cls(self.url, **self.options)
 
-        # Connections are stored as task-local state.
-        self._connection_context: ContextVar = ContextVar("connection_context")
+        # Connections are stored as task-local state, and cannot be garbage collected,
+        # since the immutable global Context stores a strong reference to each ContextVar
+        # that is created. We need these local ContextVars since two Database objects
+        # could run in the same asyncio.Task with connections to different databases.
+        self._connection_contextvar: ContextVar[Optional["Connection"]] = ContextVar(
+            f"databases:Database:{id(self)}"
+        )
 
         # When `force_rollback=True` is used, we use a single global
         # connection, within a transaction that always rolls back.
@@ -113,7 +119,7 @@ class Database:
             self._global_transaction = None
             self._global_connection = None
         else:
-            self._connection_context = ContextVar("connection_context")
+            self._connection_contextvar.set(None)
 
         await self._backend.disconnect()
         logger.info(
@@ -187,12 +193,12 @@ class Database:
         if self._global_connection is not None:
             return self._global_connection
 
-        try:
-            return self._connection_context.get()
-        except LookupError:
+        connection = self._connection_contextvar.get(default=None)
+        if connection is None:
             connection = Connection(self._backend)
-            self._connection_context.set(connection)
-            return connection
+            self._connection_contextvar.set(connection)
+
+        return connection
 
     def transaction(
         self, *, force_rollback: bool = False, **kwargs: typing.Any
@@ -344,9 +350,15 @@ class Transaction:
         self._connection_callable = connection_callable
         self._force_rollback = force_rollback
         self._extra_options = kwargs
-        self._transaction_context: ContextVar[TransactionBackend | None] = ContextVar(
-            "transaction_context"
-        )
+
+        # This ContextVar can never be garbage collected - similar to the ContextVar
+        # at Database._connection_contextvar - since the current Context has a strong
+        # reference to every ContextVar that is created. We need local ContextVars since
+        # there may be multiple (even nested) transactions in a single asyncio.Task,
+        # which each need their own unique TransactionBackend object.
+        self._transaction_contextvar: ContextVar[
+            Optional[TransactionBackend]
+        ] = ContextVar(f"databases:Transaction:{id(self)}")
 
     async def __aenter__(self) -> "Transaction":
         """
@@ -390,7 +402,11 @@ class Transaction:
     async def start(self) -> "Transaction":
         connection = self._connection_callable()
         transaction = connection._connection.transaction()
-        self._transaction_context.set(transaction)
+
+        # Cannot store returned reset token anywhere, for the same reason
+        # we need a ContextVar in the first place - `self` is not
+        # a safe object on which to store references for concurrent code.
+        self._transaction_contextvar.set(transaction)
 
         async with connection._transaction_lock:
             is_root = not connection._transaction_stack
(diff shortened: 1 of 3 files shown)


databases/core.py
```diff
@@ -344,6 +344,9 @@ def __init__(
         self._connection_callable = connection_callable
         self._force_rollback = force_rollback
         self._extra_options = kwargs
+        self._transaction_context: ContextVar[TransactionBackend | None] = ContextVar(
+            "transaction_context"
+        )
```
madkinsz, line 349: fyi https://docs.python.org/3/library/contextvars.html#contextvars.ContextVar

> Important: [Context](https://docs.python.org/3/library/contextvars.html#contextvars.Context) Variables should be created at the top module level and never in closures. Context objects hold strong references to context variables which prevents context variables from being properly garbage collected.

databases/core.py
```diff
@@ -385,31 +388,38 @@ async def wrapper(*args: typing.Any, **kwargs: typing.Any) -> typing.Any:
...
-        async with self._connection._transaction_lock:
-            is_root = not self._connection._transaction_stack
-            await self._connection.__aenter__()
-            await self._transaction.start(
-                is_root=is_root, extra_options=self._extra_options
-            )
-            self._connection._transaction_stack.append(self)
+        connection = self._connection_callable()
+        transaction = connection._connection.transaction()
+        self._transaction_context.set(transaction)
```
madkinsz, line 393: Technically you should probably stash the "token" this returns and reset the variable via the token instead of setting it to `None` later, but I'm not sure if there would ever be nesting in normal usage.

madkinsz:
Welcome and thanks for contributing! For some context, I'm a member of encode but not a maintainer of this project. Reading this out of curiosity and just have a couple notes about proper usage of the context var.

databases/core.py
zevisert: Yeah, I read that but it didn't click until coming back to review this. I added some comments to the code since the design around these contextvars is pretty opaque and foreign for a lot of people when reading it.

As the docs says, the reason that context variables _should_ be created at module level is that they automatically register with the current `contextvars.Context`, which creates a strong (aka counted) reference from the context to the new context variable. Since the Context keeps a reference to the context variable, won't be garbage collected until the Context is destroyed. Many async programs will use one long-lived asyncio.Task (containing a single Context), and even with my changes Databases will still be leaking a ContextVar with every call to `async with db.transaction()`, or the lower level `transaction = await db.transaction()`. In my view, decorator-usage (eg for a FastAPI route) will not leak these ContextVars, since only one is created per `databases.Transaction` instance, which is used (and re-ured) as the wrapping decorator. 

The problem with module level decorators for Transactions is that two transactions could happen in the same task, and these need to remain as independent instances. Take this example: 

```py
import asyncio
import databases
db = databases.Database(...)

async def nested_transactions():
    async with db.transaction():
        db.execute(...)
        async with db.transaction():
            db.execute(...)

async def main() -> None:
    # Non-concurrent example, but with multiple Tasks
    await asyncio.create_task(nested_transactions)

asyncio.run(main())
```

The above script needs to maintain two separate `databases.interfaces.TransactionBackend`  instances, one for the inner transaction, and one for the outer. We of course also need to keep the code safe for concurrent execution, so plain globals are out (due to race conditions just like if we were using threads). If these was a single module level context variable to store the transaction's backend, the nested transaction would overwrite the outer's reference. That's why I stuck with the instance level contextvar.

zevisert: Although, with writing that out, I wonder if a module level context variable that holds a task-local dictionary that maps `databases.Transaction` instances to `databases.interfaces.TransactionBackend` instances could work. We'd get one module level context variable, but still be able to store references to multiple transaction backend objects per task...

zevisert: Also, Databases currently has a non-global context variable to store the `databases.Database`'s connection instance. In similar fashion, it seems that we couldn't use a module level context variable for that either, since there could be usage like: 

```python
import asyncio
import databases

db1 = databases.Database(...)
db2 = databases.Database(...)

async def copy_something():
    something = db.execute(...)
    db.execute(process(something)

asyncio.run(main)
```

Obviously, in this example, we need two independent connection objects, one for each instance. But, perhaps the `ContextVar[dict[Database, Connection]]` type could work here too...

databases/core.py
zevisert: Yeah, that's part of the contextvars library I don't really understand. The current implementation for connections doesn't do this right now either, though. Worse, it was assigning a newly created ContextVar to clear the state of the connection.

I have to stash the token somehow, but since we're not immediately restoring the value, I think I'd need yet another context variable to keep the reset token! I can't just use an instance property, for the same reason that this PR exists, either, so where does it end? It seems that the API is designed only for cases like this: 
```python
from contextvars import ContextVar

var = ContextVar('some_var')

async def some_library_function():
    token = var.set(42)
    try: 
        await _do_something()
    finally:
        var.reset(token)
```

Since databases has separate time-of-creation and time-of-use for these context vars (eg `Transaction.start()`, versus `Transaction.commit()` or `Transaction.rollback()`), I don't see a clear way to keep the token around - which is why I opted for setting the value to None instead. Databases is not using these ContextVars to interface with other codebases, so as Databases is the sole owner, I think setting to None is fine.

zevisert:
Thanks for the review. I'm eager to move this forward. We use databases in one of our products, and when load testing this certainly fixes an issue with concurrency. Even your two comments were pretty insightful, so if you have another encode member who is interested in taking a look that would be great. 

I'll also try the module level context var solution you helped me discover in our codebase to see if that can still fix the issue. As is stands Databases was (and continues) leaking memory from it's usage of non-module level contextvars, but if I can fix that here too then that's great!

databases/core.py
madkinsz: > If these was a single module level context variable to store the transaction's backend, the nested transaction would overwrite the outer's reference. That's why I stuck with the instance level contextvar.

If you reset the variable with the token the outer context's value would be restored when the inner context exits, right?

databases/core.py
zevisert: Yeah, neither this PR nor the current master is safe in this regard. I explored the idea of a contextvar containing a dict and I have something that passes all tests now, but I need to also confirm with an MCVE as well / add some tests since the tests passed before this PR anyway.

zevisert:
@madkinsz Here's a minimal example - turns out I could get it pretty small. This fails on master, but passes perfectly fine here: 

```python
# file: tests/test_concurrent_tasks.py
import pytest
import asyncio
from tests.test_databases import async_adapter, DATABASE_URLS


@async_adapter
async def test_concurrent_tasks(database_url: str):
    """
    Test concurrent tasks.
    """

        # This could be an eg, fastapi / starlette endpoint
        @database.transaction()
        async def read_items() -> dict:
            return dict(await database.fetch_one(query="SELECT 1 AS value"))

        # Run 10 concurrent **tasks**
        tasks = (asyncio.create_task(read_items()) for _ in range(10))
        responses = await asyncio.gather(*tasks, return_exceptions=True)

        # Check the responses
        assert all(isinstance(response, dict) for response in responses)
        assert all(response["value"] == 1 for response in responses)
```

<details>
<summary>I'm using <code>docker compose</code> and <code>pytest</code> to run this</summary>

```yaml
# file: test.docker-compose.yaml
version: "3"
services:
  postgres:
    image: postgres:14
    environment:
      POSTGRES_USER: username
      POSTGRES_PASSWORD: password
      POSTGRES_DB: testsuite
    ports:
      - 5432:5432

  mysql:
    image: mysql:5.7
    environment:
      MYSQL_USER: username
      MYSQL_PASSWORD: password
      MYSQL_ROOT_PASSWORD: password
      MYSQL_DATABASE: testsuite
    ports:
      - 3306:3306
```

```console
$  docker compose --file test.docker-compose.yaml up --detach 
[+] Running 2/2
 ⠿ Container databases-mysql-1     Started
 ⠿ Container databases-postgres-1  Started   

$ git log -1 --oneline
b6eba5f (HEAD -> master, upstream/master, upstream/HEAD, github/master) Bump wheel from 0.37.1 to 0.38.1 (#524)

$ DATABASE_URLS=(
>     'sqlite:///testsuite'
>     'sqlite+aiosqlite:///testsuite'
>     'mysql://username:password@localhost:3306/testsuite'
>     'mysql+aiomysql://username:password@localhost:3306/testsuite'
>     'mysql+asyncmy://username:password@localhost:3306/testsuite'
>     'postgresql://username:password@localhost:5432/testsuite'
>     'postgresql+aiopg://username:password@127.0.0.1:5432/testsuite'
>     'postgresql+asyncpg://username:password@localhost:5432/testsuite'
> )
$ export TEST_DATABASE_URLS=$(IFS=, ; echo "${DATABASE_URLS[*]}")

$ pytest -k test_concurrent_tasks --verbose --exitfirst  # --exitfirst for brevity, they all fail or hang
==================================== test session starts =====================================
platform linux -- Python 3.10.8, pytest-7.1.2, pluggy-1.0.0 -- /opt/tooling/pyenv/versions/encode-databases/bin/python3.10
cachedir: .pytest_cache
rootdir: /code/oss/github/encode/databases
plugins: cov-3.0.0, anyio-3.6.2
collected 360 items / 352 deselected / 8 selected                                            

tests/test_concurrent_fastapi.py::test_concurrent_tasks[sqlite:///testsuite] FAILED    [ 12%]

========================================== FAILURES ==========================================
_________________________ test_concurrent_tasks[sqlite:///testsuite] _________________________

database_url = 'sqlite:///testsuite'

    @async_adapter
    async def test_concurrent_tasks(database_url: str):
        """
        Test concurrent tasks.
        """
    
            # This could be an eg, fastapi / starlette endpoint
            @database.transaction()
            async def read_items() -> dict:
                return dict(await database.fetch_one(query="SELECT 1 AS value"))
    
            # Run 10 concurrent **tasks**
            tasks = (asyncio.create_task(read_items()) for _ in range(10))
            responses = await asyncio.gather(*tasks, return_exceptions=True)
    
            # Check the responses
E >         assert all(isinstance(response, dict) for response in responses)
E           assert False
E            +  where False = all(<generator object test_concurrent_tasks.<locals>.<genexpr> at 0x7f1b26e86420>)

tests/test_concurrent_fastapi.py:25: AssertionError
================================== short test summary info ===================================
FAILED tests/test_concurrent_fastapi.py::test_concurrent_tasks[sqlite:///testsuite] - asser...
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
======================== 1 failed, 352 deselected, 1 warning in 0.28s ========================
^CException ignored in: <module 'threading' from '/opt/tooling/pyenv/versions/3.10.8/lib/python3.10/threading.py'>
Traceback (most recent call last):
  File "/opt/tooling/pyenv/versions/3.10.8/lib/python3.10/threading.py", line 1567, in _shutdown
    lock.acquire()
KeyboardInterrupt: 

$ git switch -  
Switched to branch 'fix-transaction-contextvar'
Your branch is up to date with 'github/fix-transaction-contextvar'.

$ git log -1 --oneline
3c0790d (HEAD -> fix-transaction-contextvar, github/fix-transaction-contextvar) fix: contextvar.get takes no keyword arguments

$ pytest -k test_concurrent_tasks --verbose --exitfirst
==================================== test session starts =====================================
platform linux -- Python 3.10.8, pytest-7.1.2, pluggy-1.0.0 -- /opt/tooling/pyenv/versions/encode-databases/bin/python3.10
cachedir: .pytest_cache
rootdir: /code/oss/github/encode/databases
plugins: cov-3.0.0, anyio-3.6.2
collected 360 items / 352 deselected / 8 selected                                            

tests/test_concurrent_fastapi.py::test_concurrent_tasks[sqlite:///testsuite] PASSED    [ 12%]
tests/test_concurrent_fastapi.py::test_concurrent_tasks[sqlite+aiosqlite:///testsuite] PASSED [ 25%]
tests/test_concurrent_fastapi.py::test_concurrent_tasks[mysql://username:password@localhost:3306/testsuite] PASSED [ 37%]
tests/test_concurrent_fastapi.py::test_concurrent_tasks[mysql+aiomysql://username:password@localhost:3306/testsuite] PASSED [ 50%]
tests/test_concurrent_fastapi.py::test_concurrent_tasks[mysql+asyncmy://username:password@localhost:3306/testsuite] PASSED [ 62%]
tests/test_concurrent_fastapi.py::test_concurrent_tasks[postgresql://username:password@localhost:5432/testsuite] PASSED [ 75%]
tests/test_concurrent_fastapi.py::test_concurrent_tasks[postgresql+aiopg://username:password@127.0.0.1:5432/testsuite] PASSED [ 87%]
tests/test_concurrent_fastapi.py::test_concurrent_tasks[postgresql+asyncpg://username:password@localhost:5432/testsuite] PASSED [100%]

======================= 8 passed, 352 deselected, 21 warnings in 0.68s =======================

```
</details>

zevisert:
Rebased to sign commits, all new and old tests passing, switched to module level contextvars to allow for the garbage collector to clean up unreferenced objects.

databases/core.py
zevisert: I switched to an implementation where the two contextvars each store a dictionary for the database connections or for the active transactions. We should now be safe in all of these cases: 

- The simple case where a Database is instantiated and used immediately - it gets its own connection
- A Database is instantiated and used in closures that run in the same tasks - the same connection is reused
- A Database is instantiated and used in closures that run in one or many descendant tasks - multiple connections are created, one per task.
- Multiple Databases are instantiated and used in the same task, eg via nesting `with` statements - each database instance uses and reuses a connection unique to itself
- Multiple Databases are instantiated and used in separate descendant tasks - each database instance uses its own connection
- In all the above cases, each Transaction instance is always paired with the connection used in the combination of the current task and parent database. The Transaction instance ID is used when multiple nested connections are used in the same task.
- This implementation now ensures a task-local mapping of transaction to transactionbackend, and support for nested nested transactions, including when a child transaction is run in a separate task.

databases/core.py
zevisert: Yes - it seems so. Using a token to reset a value brings back the value that was current before the contextvar.set(...) call was made - regardless of the order the tokens are used. The old value is simply stored on the token.

```python
>>> import contextvars
>>> var = contextvars.ContextVar('var', default={'id': 'default'})
>>> token0 = var.set({'id': 1})
>>> token1 = var.set({'id': 2})
>>> token2 = var.set({'id': 3})
>>> token3 = var.set({'id': 4})
>>> var.get()
{'id': 4}
>>> var.reset(token1)
>>> var.get()
{'id': 1}
>>> var.reset(token3)
>>> var.get()
{'id': 3}
>>> var.reset(token2)
>>> var.get()
{'id': 2}
>>> var.reset(token0)
>>> var.get()
{'id': 'default'}
>>> token3.old_value
{'id': 3}
```

madkinsz:
Hey! I took some time to poke around at this today. It seems a little wild to use a `ContextVar` to store a dictionary that includes the `asyncio.Task` in it (the whole point of them is that they handle task locality for you). I'm not sure what we gain by using a context variable in this case; we can just use an instance dictionary and remove a lot of complexity.

For example, I've implemented this at https://github.com/encode/databases/compare/master...madkinsz:example%2Finstance-safe and it passes the MRE you provided (I only checked SQLite because I didn't want to deal with docker-compose)

zevisert:
Yeah, this seems great. I've adopted/merged your changes into this MR to accompany the new tests I'm contributing here. I kept a few of the assertions I had added when resolving conflicts - I saw a few of them in the existing code. Let me know what encode's stance is on `assert` as well as `del some_dict[some_key]` vs `some_dict.pop(some_key)` as it seems like we have different styles there too.

Thanks for your help on this, I'm excited for sentry to finally stop notifying us about the errors from this issue!

databases/core.py
zevisert: Resolving: new implementation avoids using context variables

databases/core.py
zevisert: Resolved: @madkinsz took a new approach that I think is best here - back to instance-local storage, but without context variables, and with dictionaries keyed on the current asyncio task

databases/core.py
```diff
@@ -63,8 +62,8 @@ def __init__(
         assert issubclass(backend_cls, DatabaseBackend)
         self._backend = backend_cls(self.url, **self.options)
 
-        # Connections are stored as task-local state.
-        self._connection_context: ContextVar = ContextVar("connection_context")
+        # Connections are stored per asyncio task
+        self._connections: typing.Dict[typing.Optional[asyncio.Task], Connection] = {}
```
madkinsz, line 66: I think we can either exclude the `Optional` here _or_ exclude the assertions that `current_task` is not null. I don't have a strong preference. I omitted the assertions in my commit because they seemed onerous and `None` is perfectly fine to store in the dictionary. However, the assertion may lead to clearer behavior if someone tries to use this with a non-asyncio event loop — does this project support trio? If so we'll need to abstract current task retrieval to behave correctly.

databases/core.py
```diff
@@ -113,7 +112,10 @@ async def disconnect(self) -> None:
             self._global_transaction = None
             self._global_connection = None
         else:
-            self._connection_context = ContextVar("connection_context")
+            current_task = asyncio.current_task()
+            assert current_task is not None, "No currently running task"
+            if current_task in self._connections:
+                del self._connections[current_task]
```
madkinsz, line 118: This style of removal is fine with me! I have a minor preference for `self._connections.pop(current_task, None)` because it's shorter but I'd stick with whatever style you've seen in this repository.

madkinsz:
Glad you think it makes sense too!

I think my last concern is the event-loop compatibility bit (e.g. trio).

@tomchristie Do you think you'd be willing to give this a look soon or point me to a better reviewer?

zevisert:
I don't think Databases supports other event loop implementations like trio, curio, etc. I see an old PR #357 that was working on adding anyio support, but it's in draft still.

databases/core.py
zevisert: No one style seems more preferred here, in encode/starlette, encode/HTTPX, or encode/orm. The other three projects have about the same number of occurrences of `if x in map: del map[x]` as they do with `map.pop(x, None)`. Since neither wins out, I'm just going to leave this as is.

databases/core.py
zevisert: Yeah.. Looking into it, dropping optional would be the most correct right now. I was going by the return type annotation, but without providing a loop argument to `asyncio.current_task(...)`, this function will [use the loop actively running the task we're calling from](https://github.com/python/cpython/blob/1dad3d10714fe9611769196fde4e94bb63eb7f17/Lib/asyncio/tasks.py#L35-L39) or raise `RuntimeError: no running event loop`.

In other words, it appears that `asyncio.current_task(...)` can only return `None` if a loop argument is provided. Since we're never giving it a loop, the only thing we can get back is a `RuntimeError` or a `Task`. That's kind of unfortunate because the [type signature](https://github.com/python/typeshed/blob/0b17fafbe3a3a75ae11431ebdee72af5bf751a9c/stdlib/asyncio/tasks.pyi#L325)  doesn't have the overload to say that a return type of `None` is only possible if a loop is provided, so we still have to either ignore the Optional or assert / cast it away. 

I'm going to keep the asserts and drop the `Optional` type.

zevisert:
Trying not to be demanding your time @tomchristie - I know how open source can be brutal for that.. _but_ if you have a minute it would be great if you or someone else from encode could take a look at this. My team would really like to ship this! Thanks!

ToGoBananas:
For me, it fixes old concurrent transactions issues.
But during tests teardown, I am getting errors like 
image

Added these changes to test it more
https://github.com/ToGoBananas/databases/commit/5150e407d2dc53b7d4302101a84543142b4ebd6b

ToGoBananas:
Also, this PR brokes code with nested transactions if second transaction in the stack executed via `gather`

zevisert:
@ToGoBananas can you provide a reproducible example? Existing tests are all passing for me locally, and here in CI as well.

ToGoBananas:
```
import asyncio

from db import get_database  # just returns global instance of databases.Database (singleton)


async def run_update_from_another_task():
    results = await get_database().fetch_one("update project set name = 'test2' where id = 10 RETURNING id")
    assert results  # we can't update newly created row


async def run():
    await get_database().connect()
    await get_database().execute("delete from project where id = 10")
    async with get_database().transaction():
        await get_database().execute("insert into project(id, name) values(10, 'test')")
        results = await get_database().fetch_one("update project set name = 'test2' where id = 10 RETURNING id")
        assert results  # we can update newly created row
        await asyncio.gather(run_update_from_another_task())

asyncio.run(run())

```

zevisert:
Gotcha, thanks for the example - that helps. Here's what I ended up running to test this more, and I have what I think is a solution for you:

<details>
<summary>@ToGoBananas MVE</summary>

```python
# file: bananas.py
import asyncio
import databases
import os
from contextlib import asynccontextmanager
from typing import AsyncIterator

db = databases.Database(os.getenv("DATABASE_URL"))


@asynccontextmanager
async def setup_testbed() -> AsyncIterator[None]:
    try:
        await db.connect()
        await db.execute("CREATE TABLE IF NOT EXISTS bananas (id int, name text)")
        print(db.url)
        yield
    finally:
        await db.execute("DROP TABLE bananas")
        await db.disconnect()


async def main():
    # Wrap with DB setup and teardown
    async with setup_testbed():

        # Make sure the item doesn't exist prior to this test
        await db.execute("DELETE FROM bananas WHERE id = 1")

        # Start a transaction in this task
        async with db.transaction():

            # Create a new item
            await db.execute("INSERT INTO bananas(id, name) values(1, 'test')")

            # Show that we can update the item in this task
            await db.execute("UPDATE bananas SET name = 'main' WHERE id = 1")
            results = await db.fetch_one("SELECT * FROM bananas WHERE id = 1")
            assert results
            print("main:", dict(results._mapping.items()))

            # Try to update the item in a new task
            await asyncio.gather(run_update_from_another_task())

            # See if the child task has a side effect that is perceived here
            results = await db.fetch_one("SELECT * FROM bananas WHERE id = 1")
            print("main:", dict(results._mapping.items()))


async def run_update_from_another_task():
    await db.execute("UPDATE bananas SET name = 'gather' WHERE id = 1")
    results = await db.fetch_one("SELECT * FROM bananas WHERE id = 1")
    assert results
    print("gather:", dict(results._mapping.items()))


asyncio.run(main())
```

A few changes, obviously from what you posted:
- Your use of a singleton is no different than just creating a global `databases.Database` in this example
- You are using `UPDATE ... RETURNING`, which is a postgres-only feature, but can be implemented across the other databases supported here with a sequential `UPDATE` then `SELECT`.
- I added a setup and teardown context manager to `CREATE` the table you are showing, then clean it up when exiting.

Using the same `docker-compose` file I posted above, I have the following urls to test with, which I added to my shell then ran this:

```shell
> DATABASE_URL="postgresql://username:password@localhost:5432/testsuite" python bananas.py
postgresql://username:password@localhost:5432/testsuite
main: {'id': 1, 'name': 'main'}
Traceback (most recent call last):
  File "/code/oss/github/encode/databases/bananas.py", line 56, in <module>
    asyncio.run(main())
  File "/opt/tooling/pyenv/versions/3.10.8/lib/python3.10/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/opt/tooling/pyenv/versions/3.10.8/lib/python3.10/asyncio/base_events.py", line 649, in run_until_complete
    return future.result()
  File "/code/oss/github/encode/databases/bananas.py", line 42, in main
    await asyncio.gather(run_update_from_another_task())
  File "/code/oss/github/encode/databases/bananas.py", line 52, in run_update_from_another_task
    assert results, f"No results found: {results}"
AssertionError: No results found: None
```

Indeed, the example you posted does fail. See the discussion below, but I think this is intended behaviour. If you want to have another task influence an active transaction, you should pass the connection you'd like to use to the other task. Here's the changes I think you should make:

```diff
--- bananas.py  2023-05-19 13:09:12.193446155 -0700
+++ bananas.py"   2023-05-19 13:09:00.252633338 -0700
@@ -1,5 +1,6 @@
 import asyncio
 import databases
+import databases.core
 import os
 from contextlib import asynccontextmanager
 from typing import AsyncIterator
@@ -39,16 +40,16 @@
             print("main:", dict(results._mapping.items()))
 
             # Try to update the item in a new task
-            await asyncio.gather(run_update_from_another_task())
+            await asyncio.gather(run_update_from_another_task(db.connection()))
 
             # See if the child task has a side effect that is perceived here
             results = await db.fetch_one("SELECT * FROM bananas WHERE id = 1")
             print("main:", dict(results._mapping.items()))
 
 
-async def run_update_from_another_task():
-    await db.execute("UPDATE bananas SET name = 'gather' WHERE id = 1")
-    results = await db.fetch_one("SELECT * FROM bananas WHERE id = 1")
+async def run_update_from_another_task(connection: databases.core.Connection):
+    await connection.execute("UPDATE bananas SET name = 'gather' WHERE id = 1")
+    results = await connection.fetch_one("SELECT * FROM bananas WHERE id = 1")
     assert results, f"No results found: {results}"
     print("gather:", dict(results._mapping.items()))
```

Now, by passing the connection that is actively using the transaction you'd like to update values within, you get what I think you were expecting.

```shell
> for url in (
>     'sqlite:///testsuite'
>     'sqlite+aiosqlite:///testsuite'
>     'mysql://username:password@localhost:3306/testsuite'
>     'mysql+aiomysql://username:password@localhost:3306/testsuite'
>     'mysql+asyncmy://username:password@localhost:3306/testsuite'
>     'postgresql://username:password@localhost:5432/testsuite'
>     'postgresql+aiopg://username:password@127.0.0.1:5432/testsuite' 
>     'postgresql+asyncpg://username:password@localhost:5432/testsuite'
> ); do DATABASE_URL=$DATABASE_URL python bananas.py; done
sqlite:///testsuite
main: {'id': 1, 'name': 'main'}
gather: {'id': 1, 'name': 'gather'}
main: {'id': 1, 'name': 'gather'}
sqlite+aiosqlite:///testsuite
main: {'id': 1, 'name': 'main'}
gather: {'id': 1, 'name': 'gather'}
main: {'id': 1, 'name': 'gather'}
mysql://username:password@localhost:3306/testsuite
main: {'id': 1, 'name': 'main'}
gather: {'id': 1, 'name': 'gather'}
main: {'id': 1, 'name': 'gather'}
mysql+aiomysql://username:password@localhost:3306/testsuite
main: {'id': 1, 'name': 'main'}
gather: {'id': 1, 'name': 'gather'}
main: {'id': 1, 'name': 'gather'}
mysql+asyncmy://username:password@localhost:3306/testsuite
main: {'id': 1, 'name': 'main'}
gather: {'id': 1, 'name': 'gather'}
main: {'id': 1, 'name': 'gather'}
postgresql://username:password@localhost:5432/testsuite
main: {'id': 1, 'name': 'main'}
gather: {'id': 1, 'name': 'gather'}
main: {'id': 1, 'name': 'gather'}
postgresql+aiopg://username:password@127.0.0.1:5432/testsuite
main: {'id': 1, 'name': 'main'}
gather: {'id': 1, 'name': 'gather'}
main: {'id': 1, 'name': 'gather'}
postgresql+asyncpg://username:password@localhost:5432/testsuite
main: {'id': 1, 'name': 'main'}
gather: {'id': 1, 'name': 'gather'}
main: {'id': 1, 'name': 'gather'}
```

---

</details>

Based on the documentation for this project, I'd argue that what you're showing here was not intended behavior, and comes as a side effect of the broken concurrency model I'm trying to fix. Here's [the line I'm referencing](https://www.encode.io/databases/connections_and_transactions/#:~:text=Transaction%20blocks%20are%20managed%20as%20task%2Dlocal%20state.%20Nested%20transactions%20are%20fully%20supported%2C%20and%20are%20implemented%20using%20database%20savepoints.):
> Transaction blocks are managed as task-local state. Nested transactions are fully supported, and are implemented using database savepoints.

To _me_, this means the effects of transactions should be restricted to the task they are running in. Nested transactions, as far as I'm reading it here, just means that you can nest transaction blocks. The documentation does not state that nested transactions are expected to work across tasks. @tomchristie or @Kludex might be able to clarify. In code, I read that to mean that this is supported:

```python
async def example():
    db = databases.Database(os.getenv("DATABASE_URL"))
    async with db.transaction():
        async with db.transaction():
            ...
```

madkinsz:
Hm it does seem like tasks should inherit the state of the parent tasks if feasible. If `run_update_from_another_task` was created _outside_ of the transaction but ran during it, I would expect it to fail. If it's created from within the transaction, it probably ought to work.

zevisert:
> it does seem like tasks should inherit the state of the parent tasks if feasible.

That brings us back to context variables then. 

> If `run_update_from_another_task` was created outside of the transaction but ran during it, I would expect it to fail. If it's created from within the transaction, it probably ought to work.

Your model of how context works with tasks is slightly off I think. Context is not transferred based on where a coroutine function is defined, it's based on where it is executed. For example, the function signature for [`asyncio.create_task(coro, *, name=None, context=None)`](https://docs.python.org/3/library/asyncio-task.html?highlight=create_task#asyncio.create_task) shows that `create_task` is responsible for defining the context that the child task is run within. The default value of `None` means the task should inherit a copy of the current context, it's not based on where you defined whatever coroutine is being passed as the `coro` argument.

---

I don't mind the switch back to ContextVars, but the consequence is that _**databases**_ may need to be more clear with what is supported. The documentation for [context variables](https://docs.python.org/3/library/contextvars.html#contextvars.ContextVar) says:

> **Important**: `ContextVar`s should be created at the top module level and never in closures. [`Context`](https://docs.python.org/3/library/contextvars.html#contextvars.Context) objects hold strong references to context variables which prevents context variables from being properly garbage collected.

I understand "closure" from that warning to also include instances. I can't find much discussion related to that warning [on the PR that added that it to the docs](https://github.com/python/cpython/pull/5685) or on the [implementation itself](https://github.com/python/cpython/issues/76617), but I don't think it's memory safe to add ContextVars as instance attributes.

This is all to say that since `ContextVar`s should be declared at module level but we want them to track instance-level properties (`connection` and `transaction`), the immediate consequence I'm foreseeing is that there can only ever be one `databases.Database` instance per python program. That's because the global context variables that are used to track instance properties natively have no mechanism to keep track of which of possibly many `databases.Database` instances the connection or transaction belongs to. That circles back to an [earlier version of this PR](https://github.com/encode/databases/pull/546/commits/75969d343b505a25f6f960c1d22ac4367e45a83f#diff-0961f329ad913ad41f63a1e465a2d97b67e66ae8e5f9ddb3c6bd10293f0ed1edR38-R52) where I had context variables containing dictionaries containing connections or transactions. It's possible to manage all of this, but it was seen as too complex/ugly initially - and even then I had an `asyncio.Task` as part of the key into that mapping so I'd have to check (not certain) if that actually would work with @ToGoBananas example.

Personally, I think that the majority of _**databases**_ users are not spawning new tasks within transactions, so I'd like to solve this with documentation. The implementation we have right now is simple and maintainable, and still allows for child tasks to influence a parent transaction if you pass them the connection that's used in the parent's task (like I showed above). FWIW, the good old days of threading did not have this idea of inherited context: `threading.Local` variables did not pass copies of themselves to child threads and it was up to the programmer to pass in context explicitly.

madkinsz:
>  Context is not transferred based on where a coroutine function is defined, it's based on where it is executed.

We're definitely on the same page about this; I just wasn't clear in my comment.

I think this pull request would be a breaking change given that example. Gathering asynchronous tasks is pretty common. 

Perhaps it's possible to use context variables more cleanly. From my previous comment:

>  It seems a little wild to use a ContextVar to store a dictionary that includes the asyncio.Task in it (the whole point of them is that they handle task locality for you).

If we exclude the tasks from the context, it might make more sense. I can try taking a look at that implementation again too.

zevisert:
Cool, glad to hear it. I'll dig in a bit more as well.

I agree that this would be a breaking change as it stands right now. I didn't see that as a big deal since **_databases_** has yet to become stable (v0.7 right now), so semantic version says that any minor version change is also to be considered breaking. Considering how subtle / problematic this concurrency bug is that has been effecting my project is, I kind of anticipated a breaking change when I came here to start fixing it.  I think that some things that may have been enabled by the previous concurrency model might be removed once we land on an implementation that corrects these concurrency issues that I've been referencing above.

databases/core.py
```diff
@@ -63,8 +62,8 @@ def __init__(
         assert issubclass(backend_cls, DatabaseBackend)
         self._backend = backend_cls(self.url, **self.options)
 
-        # Connections are stored as task-local state.
-        self._connection_context: ContextVar = ContextVar("connection_context")
+        # Connections are stored per asyncio task
+        self._connections: typing.Dict[asyncio.Task, Connection] = {}
```
circlingthesun, line 66: Won't this leak memory? Is it maybe better to use a WeakKeyDictionary such that when the task is complete the key would be removed from the dictionary?

madkinsz: The connection is dropped from the dict when it is closed — it shouldn't really leak memory.

madkinsz:
@zevisert this works for both examples; I converted @ToGoBananas's example into an actual test.

https://github.com/encode/databases/commit/dccb47dc57a18fe00d801d0bcdfbf9150bc0c9ae

databases/core.py
circlingthesun, line 118: Maybe I'm missing something, but this appears to the only place where items are removed from the _connections dict. So if I'm running a web service that calls connect() when the service starts up and disconnect() when it shuts down, but in the interim I'm serving each request in a new task, I'd have an entry in the _connections dictionary for every request that used the database until my service shuts down. Also, because the current task is used as a key in the dict any request that used the database won't have it's task state garbage collected, as the dict is holding a reference to it.

zevisert: Oh I see that now too - thank you! New tasks do get allocated a new connection, and they aren't necessarily going to each call disconnect before they complete.

- Main task calls `db.connect`, a new `Connection` is created for it
- Main task creates another task 
- Child task has a new `Connection` allocated because it uses `db.execute`, or alike
- Child task does not call `db.disconnect` before it is completed
- Child task completes
- `db._connections` still has the child task as a key referencing the child's connection.

madkinsz: 👍 thanks — this should be resolved by not caching per task and instead leveraging context variables. Otherwise you're right that a weakkey dict makes sense.

databases/core.py
zevisert: Two things here:
1. If tasks raised something and they weren't using context managers to safely close resources (eg: `transaction = database.transaction(); transaction.start(); raise ValueError` is enough) then cleanup code wouldn't run, so yes this could have leaked memory. 
2. Storing the `Task` as a workaround for avoiding asyncio's context variables prevented child tasks from inheriting active connections and transactions. I've gone back to `ContextVar`s, but the potential for strong references of a dictionary keeping forgotten / failed transactions and connections in memory still exists. I switched to `WeakKeyDictionary` like you suggested and added some tests to confirm the freeing of resources that didn't call teardown code.

zevisert:
Summary of the latest changes: 
- Reverted back to `ContextVar`s to support transaction (and connection) inheritence for tasks that are started within a context where a transaction or connection is already established. Thanks to @ToGoBananas for joining in to point out the flaw with tracking by `asyncio.Task`!
- Since `ContextVar`s should be module scoped, and because we _should_ support multiple _**databases**_ instances in the same progam; we're storing a Mapping type in the `ContextVar`. This allows us to make sure we're associating the right state to the right instance. This is essentially the core of what this PR is proposing. Thank you @madkinsz for great communication and inspiration along this path.
- Since we have a global mapping of various instances to connections and transactions, we could leak memory if those items are not properly cleaned up. User code could easily do this if context managers are avoided, which is a documented approach when using _**databases**_ (see docs on low-level transaction management). So help solve for when user code does not call cleanup, we're are using `WeakKeyDictionary` as the `Mapping` type stored by the `ContextVar`s. Now the garbage collector is free to clean up unused resources in these cases, thanks to us no longer storing strong references to objects involved in concurrent operations. Thank you @circlingthesun for picking up on that so quickly!

zevisert:
Earlier I suggested that this PR be considered a breaking change - but with this latest approach, I think that isn't the case anymore. @madkinsz may be more mature in thinking about this than I am though, considering their work on [poetry-relax](https://github.com/madkinsz/poetry-relax). I read some of the supporting articles there, and generally agree that SemVer can be a bit unwieldy. I think that this PR is now _aiming_ to make no breaking changes, but I can't say for sure that these corrections to concurrency won't break someone's code.

This PR has less impact now than where it had previously gotten to, given that inheritance to child tasks is supported and working correctly now.

tests/test_databases.py
```diff
@@ -477,6 +480,212 @@ async def test_transaction_commit(database_url):
...
+
+        # If that were magically cleared, the transaction would be cleaned up,
+        # but as it stands this always causes a hang during teardown at
+        # `Database(...).disconnect()` if the transaction is not closed.
+        transaction = database.connection()._transaction_stack[-1]
+        await transaction.rollback()
+        del transaction
+
+        # Now with the transaction rolled-back, it should be cleaned up.
+        assert len(open_transactions) == 0
```
zevisert, line 686: I got a bit blocked with this test, as I noted in these comments. I am trying to test that _if a transaction is aborted without calling any teardown logic_ then the weak map in the context variable does not keep a reference to it.

`databases.core.Connection` objects have a stack (using `list`) of transactions that have been opened from them, and that stack is managed through exit handing code. If we're not calling that in this test, then there will still be a strong reference to the transaction and it won't get garbage collected. 

On one hand, I think this is programming error and it's not on _**databases**_ to account for. On the other, you could make the same argument about the use of `WeakKeyDictionary` to avoid blocking the garbage collector from cleaning up these transactions. All database drivers will hang indefinitely during test teardown `transaction.rollback()` was not called. 

I can think of two approaches here:
- Add a finalizer to `databases.core.Transaction` to call `.rollback` if nothing else has by the time they are being garbage collected. This probably requires the next approach as well though for objects to actually have no references and get gc'd..
- We could maybe use a `WeakValueDictonary` like a list (keys of incrementing integers) for the `_transaction_stack`, so that if a transaction is somehow dropped by a user, _**databases**_ can still clean up. This feels a bit "hairy" though, and I'd be more inclined to call this user-error..

madkinsz: Hm I'm not sure I entirely follow here, but it sounds bad to hang on shutdown. Is that a change in behavior?

zevisert: This is not new behaviour, I've triple-checked that. This is existing behaviour that has been exposed through these new tests that are checking the cleanup functionality of weakref-based collections. It is not introduced, changed, or fixed in this PR - the effect here is that I cannot completely test that references to `databases.core.Transaction` instances are discarded by variables going out of scope.

While this of course isn't on the happy path - scenarios where a transaction instance is deleted without calling `.commit()` or `.rollback()` are possible if users are using the [documented low-level transaction management API](https://www.encode.io/databases/connections_and_transactions/#:~:text=For%20a%20lower%2Dlevel%20transaction%20API%3A), and they didn't set up their try/except/else logic correctly (this is why I consider this user-error).

```py
# MVE that hangs on master, 0.7.0, 0.6.2, 0.5.5, 0.4.3, 0.3.2, 0.2.6, 0.1.12 
import databases, asyncio, os

async def mve():
    async with databases.Database(os.environ['TEST_DATABASE_URL']) as db:
        transaction = db.transaction()
        await transaction.start()

asyncio.run(mve())
```

It also makes sense that this hangs on shutdown. From the database driver's perspective, there is an open transaction that has not executed a `COMMIT` or `ROLLBACK` statement, better wait for that to happen before shutting down.

madkinsz: 👍 as long as it's not a regression I think it makes sense to track it in a new issue to determine the best solution. I think that ideally we'd warn that a transaction was left open rather than hanging forever.

zanieb (approved):
This looks good to me.

Can you update the documentation to note that these things are task-local but inherited by child tasks?

tests/test_databases.py
zevisert: It might not even need to be a new issue. I can try to see if there's something calling this out already, otherwise I'll create a new issue.

zevisert:
> Can you update the documentation to note that these things are task-local but inherited by child tasks?

Please take a look - feel free to edit / suggest further. I'm not sure if these updated examples are too verbose or not

zevisert:
Heading to draft for a moment - the MVE in #134 seems really closely related to this issue, but is still failing in the same way as some of the other issues I've been able to resolve with this.

zevisert:
Okay thanks to @reclosedev's issue #134, I've circled back around to thinking that users probably _don't_ want **implicit** connection inheritance across tasks, BUT do want transaction inheritance if they explicitly reuse the same connection among descendant tasks.

Their example boils down to this: 

```python
import os
import asyncio
import databases

database = databases.Database(os.environ["DB_URL"])

async def worker():
    async with database.transaction():
        await database.execute("SELECT 1 + 1")

async def main():
    await database.connect()
    await database.execute("SELECT 1 + 1")
    await asyncio.gather(*[worker() for _ in range(5)])

asyncio.run(main())
```

That is: 
1. Instantiate a `Database` and `.connect` to set up a connection pool with the database backend. (`.connect` is and has always been lazy, actual _connections_ are established by `database.connection()`)
2. Do something that interacts with the database, causing a connection to be created for the parent task. 
3. Spawn a bunch of workers to do independent work, notably not from within a transaction. 

- This sequence of actions is _very_ similar to a typical FastAPI user (or other web frameworks). They create a database instance that maybe checks for a given migration state on startup, then serves requests in their own tasks.

In the current version of this PR (and on master) this example is still broken because we are using ContextVars such that the current connection is inherited by all descendant tasks. The descendant tasks cannot do parallel work within transactions, because they all share the same connection - and a single backend connection can only support nested transactions, **not concurrent transactions**.

---

This all makes me think that the best solution is for each database instance to have an attribute that tracks connections in used per-task via something like I had been playing with in 8370299b7b8bd4ed791fbdcf2805321ea12aac9f:

```python
class Database:
    _active_connections: WeakKeyDictionary[asyncio.Task, Connection]
```

This change would make connections task-local with no inheritance. Each new task acquires a new connection from the backend's connection pool unless a connection is explicitly provided to the task. The implicit connection inheritance to descendant tasks is what seems to be the cause of most user frustration with concurrent code. #230 explicitly calls this out.

TransactionBackends, on the other hand, should still be tracked with ContextVariables so that transactions nest given the same connection is used - that's the whole point of `Connection._transaction_stack` anyway. I roughly suggested [this pattern earlier when responding](https://github.com/encode/databases/pull/546#issuecomment-1555186279) to @ToGoBananas example, as at the time we had no context variables at all so child tasks did not inherit connections, and he was looking for support for the case where a child task needed to influence the state of a transaction in the parent. My suggestion was to explicitly pass the connection with the open transaction to the descendant task. Passing the connection brings the state of any active transactions on that connection along with it.

zevisert:
Oh and another thing I found just now with the current implementation - context variables undergo a _shallow clone_ when context is copied for new tasks. That means that we had been undoing any isolation that context variables had been providing by mutating the WeakKeyDictionary if it had been created before any descendant tasks started. Simple fix though, just need to treat those dictionaries as immutable.

tests/test_databases.py
zevisert: Follow up issue: #554

zevisert:
Anything more I can do for anyone here to get this reviewed and released?

zanieb (approved):
I think this has landed in a good place. However, I am not a frequent maintainer of this repository and we'll need someone else to look it over before it can be merged.

zanieb:
Thanks for your patience!

zevisert:
Thanks for the merge! Could we get @Kludex or @aminalaee to cut a new release to that we can actually have this bug corrected in our codebases? Is there a changelog or blog article I can help write for this? Is there a release cadence or policy that my team can plan from?

zanieb:
@zevisert as far as I can tell, I can make a release. I do not think anyone else is available to maintain this project.

I'd definitely appreciate if you opened a pull request for a changelog entry at https://github.com/encode/databases/blob/master/CHANGELOG.md

There is no planned release cadence for this project. I'd like to release this change and some other fixes but there is little active development here.

PR #576, opened 2023-11-28
enabler(backends): allow ssl string parameters in PostgreSQL URL

### LINKED ISSUE #575: Allow passing string values for ssl parameter in PostgreSQL backend
When using the PostgreSQL backend, you can pass the `ssl` key in the query string, with the values `true` or `false`. These are converted to boolean and passed as arguments to [asyncpg.create_pool](https://magicstack.github.io/asyncpg/current/api/index.html#asyncpg.pool.create_pool)

The asyncpg library accepts other values than only `True` and `False`, which can be used to choose how certificate validation is done (or not done).

For the record, when setting `ssl=true`, the ssl mode used is `prefer`, which will fallback to plain if SSL connection fails, so it is not a secure default.

I'm going to send a PR that permits to send string values, but it will not change the default settings.

Exagone313:
Closes #575.

The underlying library asyncpg accepts string values in the ssl parameter. The old code only accepted the values true and false, which are converted to boolean.

### DIFF
--- a/databases/backends/postgres.py
+++ b/databases/backends/postgres.py
@@ -55,7 +55,8 @@ class PostgresBackend(DatabaseBackend):
         if max_size is not None:
             kwargs["max_size"] = int(max_size)
         if ssl is not None:
-            kwargs["ssl"] = {"true": True, "false": False}[ssl.lower()]
+            ssl = ssl.lower()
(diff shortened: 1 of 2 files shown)


zanieb (approved):
Thanks!

PR #582, opened 2024-02-21
Fix some column types being parsed twice

pmdevita:
After the SQLAlchemy 2.0 fixes, ormar's test suite highlighted a problem with JSON and enum types getting parsed from the database a second time. This should fix the problem. I'm not sure if there are any other data types affected at the moment, it should hopefully be just these two. I've also added appropriate tests for these data types.

Let me know if I need to change anything, thanks!

### DIFF
--- a/databases/backends/common/records.py
+++ b/databases/backends/common/records.py
@@ -1,6 +1,6 @@
 import enum
 import typing
-from datetime import date, datetime
+from datetime import date, datetime, time
 
 from sqlalchemy.engine.interfaces import Dialect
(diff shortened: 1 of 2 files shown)


pmdevita:
TIME also needed to be fixed, added a test for it along with DATE

tarsil (approved):

Kludex:
Feel the owner here. I trust you.

tarsil:
@pmdevita thank you for your submission here as it was a nice catch from `ormar`. So far the changes you have applied, it seems like its covering the scenarios properly.

PR #587, opened 2024-03-05
Use correct type hints for query methods

ghost:
What title says.


ghost:
Is there a reason to have RecordInterface at all? It declares only two methods omitting lots of other methods and magic methods in common.Record. If someone needs a custom Record they should extend it, not implement RecordInterface.

